PR #28, opened 2018-02-05
switch to solcjs, fix formatting unit test, update chalk dependency

linusnorton:
Unfortunately I also had to update chalk to get it to build correctly.

### DIFF
--- a/lib/abiParser.ts
+++ b/lib/abiParser.ts
@@ -1,7 +1,9 @@
 import debug from "./debug";
 import { EvmType, VoidType, parseEvmType } from "./typeParser";
-import { yellow } from "chalk";
+import chalk from "chalk";
 import { MalformedAbiError } from "./errors";
+
+const { yellow } = chalk;
 
 export interface AbiParameter {
   name: string;
--- a/lib/cli.ts
+++ b/lib/cli.ts
@@ -1,6 +1,6 @@
 #!/usr/bin/env node
 import { readFileSync, writeFileSync } from "fs";
-import { blue, red, green } from "chalk";
+import chalk from "chalk";
 import { join, dirname, parse, relative } from "path";
 import { pathExistsSync } from "fs-extra";
 import * as glob from "glob";
@@ -11,6 +11,7 @@ import { parseArgs } from "./parseArgs";
 import { copyRuntime } from "./copyRuntime";
 import { extractAbi } from "./abiParser";
 
+const { blue, red, green } = chalk;
 const cwd = process.cwd();
 
 async function main() {
--- a/package.json
+++ b/package.json
@@ -31,12 +31,14 @@
     "web3-typescript-typings": "^0.7.2"
   },
   "dependencies": {
-    "chalk": "^2.1.0",
+    "bignumber.js": "^5.0.0",
+    "chalk": "^2.3.0",
     "command-line-args": "^4.0.7",
     "debug": "^3.0.1",
     "fs-extra": "^4.0.2",
     "glob": "^7.1.2",
-    "prettier": "^1.9.1"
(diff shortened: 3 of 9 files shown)


lib/abiParser.ts
```diff
@@ -1,8 +1,10 @@
 import debug from "./debug";
 import { EvmType, VoidType, parseEvmType } from "./typeParser";
-import { yellow } from "chalk";
+import chalk from "chalk";
 import { MalformedAbiError } from "./errors";
 
+const { yellow } = chalk;
```
krzkaczor: please revert. Why do you prefer object destruction instead doing this on one go while importing?

lib/cli.ts
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env node
 import { readFileSync, writeFileSync } from "fs";
-import { blue, red, green } from "chalk";
+import chalk from "chalk";
```
krzkaczor: same

lib/cli.ts
```diff
@@ -11,6 +11,7 @@ import { parseArgs } from "./parseArgs";
 import { copyRuntime } from "./copyRuntime";
 import { extractAbi } from "./abiParser";
 
+const { blue, red, green } = chalk;
```
krzkaczor: same

package.json
```diff
@@ -31,12 +31,14 @@
     "web3-typescript-typings": "^0.7.2"
   },
   "dependencies": {
-    "chalk": "^2.1.0",
+    "bignumber.js": "^5.0.0",
```
krzkaczor: bignumber should be peer dependency

test/integration/ContractWithOverloads.spec.ts
```diff
@@ -2,7 +2,7 @@ import { expect } from "chai";
 import { deployContract } from "./utils/web3Contracts";
 
 import { web3 } from "./web3";
-import { ContractWithOverloads } from "./abis/ContractWithOverloads";
+import { __ContractWithOverloads_sol_ContractWithOverloads as ContractWithOverloads } from "./abis/__ContractWithOverloads_sol_ContractWithOverloads";
```
krzkaczor: isn't there a way to make solcjs generate more readable file names?

krzkaczor: It pollutes a lot of code otherwise

lib/abiParser.ts
linusnorton: Unfortunately that format is no longer supported, see https://github.com/chalk/chalk/issues/215#issuecomment-338889272

test/integration/ContractWithOverloads.spec.ts
linusnorton: I didn't find one. I `cd` into that directory to stop it generating even more noise. I will dig a bit more and get back to you.

lib/abiParser.ts
krzkaczor: Uhh, that's horrifying but I guess we need to roll with it 👍

linusnorton: Yeah, sorry about that. The maintainer of chalk seems quite adamant that it is the best way to do it :/

package.json
linusnorton: I think it should be both as the project needs it and downstream projects need. Is that how peerDependencies work? You've got web3 as both a dependency and a peerDependency so I guess so.

linusnorton:
@krzkaczor I think this is as ready as it's ever going to be. I didn't find a way to tidy up the contract filenames (apart from manually renaming them).

krzkaczor:
@linusnorton thanks! Sorry, i was busy. I will make a review of this and the other branch before monday!

PR #33, opened 2018-03-13
Disable tslint in typechain-runtime.ts

lukeautry:
- Typechain-runtime gets copied to the project
- If the linting rules don't match, you'd have to go into this file and add a tslint disable tag every time it's generated.

### DIFF
--- a/runtime/typechain-runtime.ts
+++ b/runtime/typechain-runtime.ts
@@ -1,3 +1,4 @@
+/* tslint:disable */
 import { BigNumber } from "bignumber.js";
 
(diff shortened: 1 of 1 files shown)


krzkaczor:
I am gonna release new version today. Fix would be even cooler if tslint disable comment would be added only when runtime gets copied. Thanks to this we could still have tslint checks during TypeChain development.

krzkaczor:
@lukeautry I will add it internal todo list ;) btw. released to npm as `0.1.3`.

PR #35, opened 2018-04-03
Added Dockerfile

jpeletier:


Dockerfile
```diff
@@ -0,0 +1,20 @@
+FROM node:9.10.1-alpine as builder
+RUN apk update && apk upgrade && \
+    apk add --no-cache bash git openssh alpine-sdk python python-dev py-pip build-base
+RUN npm install -g yarn
+ADD package.json .
```
krzkaczor: lockfile is missing

Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+RUN npm install
```
krzkaczor: you should always use yarn for reproducible builds

Dockerfile
jpeletier: Are you referring to `package-lock.json`? it is not in the repository. If you are referring to the yarn lock file, it is not necessary at that stage since the objective is only to install the npm dependencies and cache these in a Docker layer.

krzkaczor: I am referring to `yarn.lock`. 

> it is not necessary at that stage since the objective is only to install the npm dependencies and cache these in a Docker layer

This is exactly why it's needed. Without this when you install npm dependencies you can't be sure what will be installed. We ran it in travis as well https://github.com/Neufund/TypeChain/blob/master/.travis.yml#L10

Dockerfile
krzkaczor: Here you should install deps with yarn so lockfile is properly utilized. I don't care really tool used to running package.json tasks

krzkaczor: btw. here i am referring to "build" as:
- install dependencies 
- transpile sources etc.

The first phase can fail if you don't use a lockfile.

jpeletier:
Please check now. I had to put the `--ignore-scripts` flag on `yarn install` to avoid building right after dependencies are installed. For some reason `yarn install` downloads dependencies AND builds right after, which is undesired within Docker context.

krzkaczor:
@jpeletier I think this is b/c of prepublish script being called (obviously) not before publishing but after installing deps 😆

krzkaczor:
Works fine! I am gonna merge it! 

The bad thing is that it's gonna be missing on master branch. Now next and master are diverging more and more (I also push some bugfixes only on master). 

I would strongly suggest you to create similar PR to master branch. I was worried that `next` branch is gonna couse problems but I didnt thought that this will happen so quickly. To support multiple web3 version we need to have one codebase with multiple generation targets.

jpeletier:
Most things in Ethereum are beta, so I wouldn't worry at this time to support old web3js versions which are going to be forgotten very soon. The ecosystem is moving very quickly. My suggestion would be that you tag `master` at this point so people can still use it as is with a deprecation notice, and move on keeping in sync with web3js as it evolves. This will allow everybody to focus all the energy on the new stuff which is where we're going anyway.

PR #40, opened 2018-04-18
Fix for https://github.com/Neufund/TypeChain/issues/39

sdesmond46:
The fix here is simply to check what type the input argument is and special case arrays. Instead of just calling to string on the array, we map the values of the array to their string representation. I added an integration test for the case as well


krzkaczor:
Great work! @sdesmond46 thanks for the fix. I have no idea why travis CI didnt run :/ let me test it manually.

PR #42, opened 2018-04-20
Fix for https://github.com/Neufund/TypeChain/issues/41 - Handle bytes as an input type

sdesmond46:
This change introduces a new `generateCodeForInputConversion` on the EvmType class which is used for converting from the input type to the type which web3 is expecting.


krzkaczor:
Thanks for another really good PR. I don't have too much experience while working with bytes array in solidity/web3 but the reasoning here seems good.

PR #43, opened 2018-04-24
Add Support for Events

sdesmond46:
This change builds of the work started in [this branch](https://github.com/Neufund/TypeChain/tree/kk/event-support) to add support for events to TypeDoc. In the generated classes, Events are member functions which can be called with filters, then either watched for all incoming events or used to get historical events.

The testing has mostly been from the integration tests I added, as well as using the modified library in my own project.


vshjxyz:
Looking forward for this! let me know if we can help to get these changes through 👍

package.json
```diff
@@ -22,7 +22,7 @@
     "chai-as-promised": "^7.1.1",
     "ganache-cli": "^6.0.3",
     "mocha": "^4.0.1",
-    "ts-node": "^3.3.0",
+    "ts-node": "3.3.0",
```
krzkaczor: is there any particular reason for updating this dependency?

krzkaczor: It makes harder to review changes (lots of package.lock changes) so I prefer to do it in separate PR.

krzkaczor:
@sdesmond46 Thanks for your hard work! It was merged with https://github.com/Neufund/TypeChain/pull/49

PR #52, opened 2018-06-09
improve check for empty contract

### LINKED ISSUE #50: createAndValidate passes for non-existent contract with MetaMask
When the contract does not exist at the specified address the `web3` instance from MetaMask will return `0x` and not `0x0` for the call to `web3.eth.getCode`. So `createAndValidate` does not throw an error. The check [here](https://github.com/Neufund/TypeChain/blob/ef4468caf20d5f5eaed02b1e4bbe90cc579b49e2/lib/generateSource.ts#L38) checks for `0x0`.

I was not able to verify what the expected behaviour of `web3` is. Will research another day.

krzkaczor:
Closes #50

### DIFF
--- a/lib/generateSource.ts
+++ b/lib/generateSource.ts
@@ -35,7 +35,7 @@ export class ${typeName} extends TypeChainContract {
     static async createAndValidate(web3: any, address: string | BigNumber): Promise<${typeName}> {
       const contract = new ${typeName}(web3, address);
       const code = await promisify(web3.eth.getCode, [address]);
-      if (code === "0x0") {
+      if (code.length < 4) {
         throw new Error(\`Contract at \${address} doesn't exist!\`);
       }
       return contract; 


Levino:
When will this be released? What is the release cycle like anyhow?

lib/generateSource.ts
```diff
@@ -35,7 +35,7 @@ export class ${typeName} extends TypeChainContract {
     static async createAndValidate(web3: any, address: string | BigNumber): Promise<${typeName}> {
       const contract = new ${typeName}(web3, address);
       const code = await promisify(web3.eth.getCode, [address]);
-      if (code === "0x0") {
+      if (code.length < 4) {
```
Levino: I would not accept this PR without a comment on what is the reasoning behind it. The comment should explain all possible cases of this, why anything longer than 3 chars is a correct smart contract and possibly link / reference the issue #50 here on github for people to understand in the future why checking for `0x0` is not sufficient. In an ideal world there would even be a regression test with metamasks `web3` to make sure nobody changes this again to the behaviour before. This was a **very dangerous** bug, because when the check passes on an non-existent contract, one will potentially call some methods anyhow and waste real ether.

krzkaczor: @Levino I also wanted to leave a comment here but then realized that it's generated code and somehow didn't want to do so. 

It's not possible to have valid smart contract with 3 characters - that's my reasoning for this check. And i think it covers various cases like `0x0` or `0x` or something even else that we are not aware of in a better way.

krzkaczor:
@Levino it's released as `0.2.0`

lib/generateSource.ts
krzkaczor: @Levino comment about empty smart contract check was added by https://github.com/Neufund/TypeChain/pull/55. 

Currently, test verifying this behavior is non-existing, we are accepting PRs :)

PR #56, opened 2018-06-11
Allow "tuple" type in ABIs

gkaemmer:
Thanks for this great library! I have an ABI that contains a output with type `tuple`, which (at least with truffle) means the type will include a `components` key with a list of sub-types.

This allows TypeChain to build typescript tuples out of those ABI tuples.

### DIFF
--- a/lib/abiParser.ts
+++ b/lib/abiParser.ts
@@ -41,6 +41,7 @@ export interface Contract {
 export interface RawAbiParameter {
   name: string;
   type: string;
+  components?: RawAbiParameter[];
 }
 
 export interface RawAbiDefinition {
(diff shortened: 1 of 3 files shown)


gkaemmer:
Hmm--I might close this for now because the ABI decoder in web3.js still doesn't support tuples, so integration tests are failing.

I mostly just wanted to prevent an error from a contract that had a `public` field with a nested struct. The auto-generated getter can't be parsed by web3 anyway, so my solution for now is just to remove the `public` modifier on the field.

krzkaczor:
Ahh, i get you. Could you commit the failing test anyway? In future, someone could continue working on it.

gkaemmer:
Sure thing--it's pushed here: https://github.com/gkaemmer/TypeChain/commit/ad930b0149ad5b8e9c186bf0e7a6c762ecf28374

krzkaczor:
That's fine. I will try to pick up from where you left when i have time. I wonder, maybe new web3 (1.0) supports it. We need to support it anyway one day so i want to introduce `target` parameter or something like this.

PR #66, opened 2018-07-29
Add estimateGas to DeferredTransactionWrapper

myspivey:
It would be great to have a way to call estimateGas through this interface. Ideally this would be called in the send process but for now this could be a good start to assist people pass more specific gas estimates instead of standard gas limits per the example.


krzkaczor:
Thanks @myspivey, this looks good! 

Could you run `yarn test:fix` to fix autofixable problems? (linting failed)

krzkaczor:
Closing in favor of https://github.com/krzkaczor/TypeChain/pull/70 Thanks!

PR #68, opened 2018-08-06
Remove unused imports with generated contracts

### LINKED ISSUE #48: remove unused imports with generated contracts
Result:
tsconfig with noUnusedLocals is not compiling because of IPayableTxParams and ITxParams are not used sometimes but they present inside the compiled source.

Expected: 
ITxParams and IPayableTxParams to not present inside build

Really good work !! Thank you for this repository :)

Regards,
Kristiyan Tachev

desfero:
Fix #48


package.json
```diff
@@ -12,7 +12,6 @@
   "license": "MIT",
   "repository": "https://github.com/Neufund/Typechain",
   "devDependencies": {
-    "@types/bignumber.js": "^4.0.3",
```
desfero: @krzkaczor do we need typings for BigNumber when they are provided internally by `bignumber.js` through `types` package.json field?

package.json
```diff
@@ -40,7 +39,6 @@
     "web3-typescript-typings": "^0.7.2"
   },
   "dependencies": {
-    "bignumber.js": "^5.0.0",
```
desfero: @krzkaczor no needed? It's already in `devDependencies` and `peerDependencies`

lib/generateSource.ts
```diff
@@ -15,16 +25,17 @@ export function generateSource(abi: Array<RawAbiDefinition>, context: IContext):
...
+
+  return `
+/* GENERATED BY TYPECHAIN VER. ${getVersion()} */
 /* tslint:disable */
   
 import { BigNumber } from "bignumber.js";
-import { TypeChainContract, promisify, ITxParams, IPayableTxParams, DeferredTransactionWrapper, DeferredEventWrapper } from '${
-    context.relativeRuntimePath
-  }'
+import * as ${runtimeNamespace} from "${context.relativeRuntimePath}"
```
desfero: @krzkaczor I'm not a fan of importing entire module thought `* as ...` but I found it the easiest approach to fix a bug.

package.json
krzkaczor: If it still compiles then probably not :D I don't remember the original reason to add it.

package.json
krzkaczor: Hmm yeah it should not be both `peerDependecies` and `dependencies`.

lib/generateSource.ts
krzkaczor: Yeah, I like this solution 👍

krzkaczor:
Nice work @desfero. I think that using `*` import is very clean solution to this problem.

PR #96, opened 2018-09-17
Kk/fix multiple files web3 1.0.0

krzkaczor:


krzkaczor:
Thanks! I added some more tests and published as `typechain@0.3.5`

PR #102, opened 2018-09-25
Ethers.js  integration and testing

kpulkit29:


krzkaczor:
Please remove `package.lock` we use `yarn`. CI fails, have you tried running `yarn test:fix:` to fix all autofixable problems?

package.json
```diff
@@ -51,6 +53,7 @@
   "dependencies": {
     "command-line-args": "^4.0.7",
     "debug": "^3.0.1",
+    "ethers": "^3.0.29",
```
krzkaczor: this should be dev dependency

package.json
```diff
@@ -28,6 +28,8 @@
     "chai": "^4.1.2",
     "chai-as-promised": "^7.1.1",
     "coveralls": "^3.0.2",
+    "ethers": "*",
```
krzkaczor: this is not needed

kpulkit29:
Please review

PR #107, opened 2018-10-05
support for tuples in arguments, return types and events - also added…

### LINKED ISSUE #105: Parse the tuple/struct type
ABIs can contain functions that take as argument or return tuples (Solidity structs). Granted, I don't think these are yet accessible from web3, but in anticipation of future updates, it would be nice if the parser didn't choke on the following contract code:
```solidity
struct Struct1 {
  uint number;
}

struct Struct2 {
  Struct1 s;
}

Struct2 public s;
```

yarrumretep:
… named arguments in return values for the web3 type definitions.

fixes #105


krzkaczor:
Thanks for your contribution @yarrumretep! I didn't have time to read PR in details but are u aware why `yarn.lock` was changed? What version of yarn are you running? This seems like `yarn.lock` format was slightly changed (new field for integrity?)

yarrumretep:
Hi @krzkaczor - I'm using 1.10.1.  Is that messing things up?  I can take a look at improving the test coverage.  Also, I'm not sure this does the right thing for either `legacy` or `truffle` - but it looks ok for the `web3-1.0.0` generator.  Truffle is not using the beta.36 version of web3-1.0.0 so it doesn't have tuple support yet.  I don't know if legacy supports tuples or not.

lib/parser/typeParser.ts
```diff
@@ -72,6 +81,9 @@ export function parseEvmType(rawType: string): EvmType {
       return new BytesType(1);
     case "bytes":
       return new ArrayType(new BytesType(1));
+    case "tuple":
+      if (!components) throw new Error("Tuple specified without components!");
```
krzkaczor: just use invariant lib as we do in other places of the codebase. You will have to use components! in next line to let compiler know that we handled it

krzkaczor:
Nice PR, I left 1 comm! Please remove any changes to yarn lockfiles since I can't review them easily. I will update it in separate PR. Thanks!

lib/parser/typeParser.ts
yarrumretep: @krzkaczor - sorry I grep'd for invariant and couldn't find any references to that library in the project.  Would you like me to include it?

yarrumretep:
Tried to eliminate the yarn.lock changes - the PR diff report is a bit wonky, but if you look at the yarn.lock diffs, there are none...  That seem ok to you?  I essentially just pulled from your master branch and committed that file on my fork.  Was surprised the PR diff still reported +590 lines...

krzkaczor:
@yarrumretep there were 2 `yarn.lock`s in your pr. There is still one not changed — thus DIFF.

lib/parser/typeParser.ts
krzkaczor: Sorry, I was mistaken that we already use it in this project. Okay don't worry about it ;) sorry about the confusion, your solution is fine anyway.

yarrumretep:
Right - there are two yarn.lock commits - one with the changes and the second backing them out.  Is that OK or is there some better way to resolve this?

krzkaczor:
@yarrumretep no, there were literally 2 lock files. https://github.com/ethereum-ts/TypeChain/pull/107/files#diff-6848d064fc673ba0b303965372931bf0 one of each is still part of your PR. Second lock file is used in tests.

yarrumretep:
Doh - my bad!  How did I miss that - got it now.  Also backed out an unnecessary gas limit increase.

krzkaczor:
Great, thanks @yarrumretep!

PR #121, opened 2018-11-06
Fix topics types

GuggerJoel:
We should be able to pass null in topics, otherwise we cannot skip some topic. This is important because topics are order-dependent.

https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_newfilter

Btw thanks for this amazing tool!


krzkaczor:
Thanks for your contibution @GuggerJoel! Fix is released as `typechain@0.3.9`

Issue #109, opened 2018-10-11
[Web3 1.0.0] Transactions (.send) should have current return type


krzkaczor:
`send` methods in web3 target have incorrect return types. 

Here is a failing test proving what's wrong: https://github.com/ethereum-ts/TypeChain/pull/129 Current return type is T (boolean in that case) where it should be a transaction object. 

This could be an error in web3 typings, not necessary in Typechain but it will probably require some changes in typechain as well.

Levino:
Sorry to say: Really bad issue. Not actionable. I know you are the maintainer, but still, come on.

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 100.0 DAI (100.0 USD @ $1.0/DAI)  attached to it as part of the Ethereum Foundation fund.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/ethereum-ts/TypeChain/109/1968).
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/ethereum-ts/TypeChain/109/1968).
* Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
* $45,943.48 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

vikaskyadav:
@krzkaczor I have a question here. Why do we need to return void in place of Task? As per my understanding, while returning void, it will be difficult to determine whether the task has finished successfully or not.

vikaskyadav:
> Has been fixed.

where is this fixed? It's PR hasn't been referenced on this issue/

krzkaczor:
nah, it's not fixed. @Levino what do you mean?

Levino:
I mean that my comment does not apply any more because the issue description has been fixed.

krzkaczor:
Ahh okay. Yeah, so it was usable before it's just a little bit too permisive. Sorry for the confusion. 

@vikaskyadav let me know if you're interested in working on this issue. I suspect its 1 line code change.

vikaskyadav:
@krzkaczor yeah. I am interested but will have to look in the tests part.

vikaskyadav:
[...]

PR #129, opened 2018-12-14
add failing test case for issue #109

krzkaczor:


Levino:
When I run the test script I get this file generated for the DumbContract (targets/web3-1.0.0/types/web3-contracts/DumbContract.d.ts).
```typescript
/* Generated by ts-generator ver. 0.0.8 */
/* tslint:disable */

import Contract, { CustomOptions, contractOptions } from "web3/eth/contract";
import { TransactionObject, BlockType } from "web3/eth/types";
import { Callback, EventLog } from "web3/types";
import { EventEmitter } from "events";
import { Provider } from "web3/providers";

export class DumbContract {
  constructor(jsonInterface: any[], address?: string, options?: CustomOptions);
  _address: string;
  options: contractOptions;
  methods: {
    returnSigned(offset: number | string): TransactionObject<string>;

    callWithBoolean(boolParam: boolean): TransactionObject<boolean>;

    callWithArray2(arrayParam: (number | string)[]): TransactionObject<(string)[]>;

    testAddress(a: string): TransactionObject<string>;

    testString(a: string): TransactionObject<string>;

    counterArray(arg0: number | string): TransactionObject<string>;

    returnAll(): TransactionObject<{
      0: string;
      1: string;
    }>;

    counterWithOffset(offset: number | string): TransactionObject<string>;

    countupForEther(): TransactionObject<void>;

    twoUnnamedArgs(
      arg0: number | string,
      arg1: number | string,
      ret: number | string,
    ): TransactionObject<string>;

    callWithBytes(byteParam: string | number[]): TransactionObject<string>;

    countup(offset: number | string): TransactionObject<void>;

    callWithArray(arrayParam: (number | string)[]): TransactionObject<string>;

    arrayParamLength(): TransactionObject<string>;
    someAddress(): TransactionObject<string>;
    counter(): TransactionObject<string>;
    SOME_VALUE(): TransactionObject<boolean>;
    byteArray(): TransactionObject<string>;
  };
  deploy(options: { data: string; arguments: any[] }): TransactionObject<Contract>;
  events: {
    Deposit(
      options?: {
        filter?: object;
        fromBlock?: BlockType;
        topics?: (null | string)[];
      },
      cb?: Callback<EventLog>,
    ): EventEmitter;

    allEvents: (
      options?: {
        filter?: object;
        fromBlock?: BlockType;
        topics?: (null | string)[];
      },
      cb?: Callback<EventLog>,
    ) => EventEmitter;
  };
  getPastEvents(
    event: string,
    options?: {
      filter?: object;
      fromBlock?: BlockType;
      toBlock?: BlockType;
      topics?: (null | string)[];
    },
    cb?: Callback<EventLog[]>,
  ): Promise<EventLog[]>;
  setProvider(provider: Provider): void;
}
```
As you can see the type with which the `TransactionObject` resolves on `send` is hardcoded to `boolean` here. There is nothing wrong with the upstream `@types/web3`. I think the issue is rather that the type `boolean` here is inferred from the return type of the solidity member method of the contract: https://github.com/ethereum-ts/TypeChain/blob/01af3109a04ca589bf5bcbb5d21dd371e065bc9f/test/integration/contracts/DumbContract.sol#L65

I think the typescript code generation of TypeChain is the culprit here. The generated file should rather look like this:

```typescript
/* Generated by ts-generator ver. 0.0.8 */
/* tslint:disable */

import Contract, { CustomOptions, contractOptions } from "web3/eth/contract";
import { TransactionObject, BlockType } from "web3/eth/types";
import { Callback, EventLog } from "web3/types";
import { EventEmitter } from "events";
import { Provider } from "web3/providers";

export class DumbContract {
  constructor(jsonInterface: any[], address?: string, options?: CustomOptions);
  _address: string;
  options: contractOptions;
  methods: {
    returnSigned(offset: number | string): TransactionObject<TransactionReceipt>;

    callWithBoolean(boolParam: boolean): TransactionObject<TransactionReceipt>;
...
```

Levino:
Look also here: https://github.com/ethereum-ts/TypeChain/blob/01af3109a04ca589bf5bcbb5d21dd371e065bc9f/lib/targets/web3/generation.ts#L78-L80

krzkaczor:
@Levino to get everyone on the same page, here's TransactionObject from web3 typings:

```typescript
export interface TransactionObject<T> {
	arguments: any[];
	call(tx?: Tx): Promise<T>;
	send(tx?: Tx): PromiEvent<T>;
	estimateGas(tx?: Tx): Promise<number>;
	encodeABI(): string;
}
```

Note that each function in abi can be either called (client side executed given current blockchain state) or sent (created TX and made persistent on blockchain). Given this definition, and these requrements what you said `TransactionObject<TransactionReceipt>` wouldn't make sense since `call` would return `Promise<TransactionReceipt>` which is obviosuly wrong since it should return what blockchain function returns. 

Furthermore, `send` should **never** return anything different that `TransactionReceipt` that's why I was proposing to change it to:


```typescript
export interface TransactionObject<T> {
	arguments: any[];
	call(tx?: Tx): Promise<T>;
	send(tx?: Tx): PromiEvent<TransactionReceipt>;
	estimateGas(tx?: Tx): Promise<number>;
	encodeABI(): string;
}
```

What do you think about this?

Levino:
1. Currently the type definitions for the transaction object are generic. The code generation of TypeChain sets the generic to the wrong type, namely the return value of the solidity method while it should always be `TransactionReceipt`. First fix your issue in your own code, then look at the upstream repo. `@types/web3` might not be to your liking, but it does not have a bug. You are using it incorrectly.
1. > Furthermore, send should never return anything different that TransactionReceipt that's why I was proposing to change it to:

    While this might be your opinion, I recommend [reading the docs](https://web3js.readthedocs.io/en/1.0/web3-eth-contract.html#id22):

    > PromiEvent: A promise combined event emitter. Will be resolved when the transaction receipt is available, OR if this send() is called from a someContract.deploy(), then the **promise will resolve with the new contract instance**.

Levino:
Also feel free to comment on the [related PR](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/31629#issuecomment-450640318) in `@types/web3`

Levino:
> Did you even read what I wrote? This would break usage of call.

I overlooked part of your statement, sorry. I agree that there is a bug in the type definitions (which I never doubted actually, but instead of focusing on the bug, people here were talking about the fix, which one had to reverse engineer the bug from). So basically it boils down to the fact that currently the following is not possible:

```typescript
const myTransaction: TransactionObject
const result: boolean = await myTransaction.call()
const receipt: TransactionReceipt = await myTransaction.send()
```
Please make a PR to `@types/web3` fixing this if you have the time. Happy to review and accept.

I have two other questions kind of off topic:
1. Should it not be possible to overwrite the type definitions locally with module augmentation instead of this "import part of the type defs and then compose them differently" approach you take? I mean the file structure of the `@types/web3` package **is not part of the interface** and can change at any time.
1. Versioning of the `@types` packages is not a strict as it is for normal npm packages. I am not sure that if a PR is merged, chaning the typ defs there, actually a version is bumped. How do you make sure the consumers of TypeChain actually always have exactly the `@types/web3` you expect? Is the whole lot hardcoded into the TypeChain output?

PR #130, opened 2018-12-17
[legacy] correctly pass boolean arguments in methods

### LINKED ISSUE #120: [legacy] invalid passing of bool contract method arguments
## TL;DR
With the `legacy` target for web3-0.2.x, `bool` arguments in contract methods are passed through a `.toString()` in the generated TS class, which results in the argument always being set to `true` on the blockchain side. 

Did not test with other targets.

### Contract
```solidity
contract TestContract {
    bool public field;
    event FieldSet(bool value);

    function setField(bool _field) public {
        field = _field;
        emit FieldSet(field);
    }
}
```

### Test
```ts
import { artifacts, contract, web3 } from './truffle-globals';
import { TestContract } from '@contracts/TestContract';
import { expect } from 'chai';
import { it } from 'mocha';

contract('TestContract', accounts => {
    let testInstance: TestContract;
    before(async () => {
        const truffleDeployedContract = await artifacts
            .require(TestContract.name)
            .new({ from: accounts[0] });
        testInstance = await TestContract.createAndValidate(
            web3,
            truffleDeployedContract.address,
        );
    });

    it('tests setting the field', async () => {
        expect(await testInstance.field).to.eq(false);
        await testInstance.setFieldTx(true).send({ from: accounts[0] });
        expect(await testInstance.field).to.eq(true);
    });

    it('tests unsetting the field', async () => {
        expect(await testInstance.field).to.eq(true);
        await testInstance.setFieldTx(false).send({ from: accounts[0] });
        expect(await testInstance.field).to.eq(false);
    });
});

```

### Test output
Most important detail: during the second test the emitted event says `value: true`, even though the method was called with `false`:
```
  Contract: TestContract
    ✓ tests setting the field (68ms)
    1) tests unsetting the field

    Events emitted during test:
    ---------------------------

    FieldSet(value: true)

    ---------------------------


  1 passing (264ms)
  1 failing

  1) Contract: TestContract
       tests unsetting the field:

      AssertionError: expected true to equal false
      + expected - actual

      -true
      +false
      
      at Object.<anonymous> (test/asdf.spec.ts:27:45)
      at Generator.next (<anonymous>)
      at fulfilled (test/asdf.spec.ts:4:58)
      at <anonymous>
      at process._tickCallback (internal/process/next_tick.js:182:7)
```

### Culprit
The `_field.toString()` argument in `TestContract.ts`.
When the `.toString()` conversion is removed manually, the test passes as expected.
```ts
  public setFieldTx(
    _field: boolean
  ): TC.DeferredTransactionWrapper<TC.ITxParams> {
    return new TC.DeferredTransactionWrapper<TC.ITxParams>(this, "setField", [
      _field.toString()
    ]);
  }
```

### env
`typechain 0.3.8`
`web3 0.20.7`
`ganache-cli 6.1.8`
`truffle 4.1.13`

quezak:
Fixes #120 

 * updated code generation for passing boolean arguments in methods
 * updated code generation for passing arrays, so nested boolean arguments work too
 * updated tests for the legacy target, to verify that boolean values are passed correctly

### DIFF
--- a/scripts/test.sh
+++ b/scripts/test.sh
@@ -1,6 +1,5 @@
 #!/usr/bin/env bash
 set -e
-set -x
 cd "$(dirname "$0")"
 cd ..
 
--- a/scripts/test.sh
+++ b/scripts/test.sh
@@ -1,5 +1,6 @@
 #!/usr/bin/env bash
 set -e
+set -x
 cd "$(dirname "$0")"
 cd ..
 
--- a/__snapshots__/DumbContract.spec.ts.js
+++ b/__snapshots__/DumbContract.spec.ts.js
@@ -226,7 +226,7 @@ export class DumbContract extends TC.TypeChainContract {
   }
 
   public callWithBoolean(boolParam: boolean): Promise<boolean> {
-    return TC.promisify(this.rawWeb3Contract.callWithBoolean, [boolParam.toString()]);
+    return TC.promisify(this.rawWeb3Contract.callWithBoolean, [boolParam]);
   }
 
   public callWithArray2(arrayParam: BigNumber[]): Promise<BigNumber[]> {
--- a/__snapshots__/DumbContract.spec.ts.js
+++ b/__snapshots__/DumbContract.spec.ts.js
@@ -114,6 +114,15 @@ export class DumbContract extends TC.TypeChainContract {
       },
       {
         constant: true,
+        inputs: [{ name: "boolArrayParam", type: "bool[]" }],
+        name: "callWithBooleanArray",
+        outputs: [{ name: "", type: "bool[]" }],
+        payable: false,
+        stateMutability: "pure",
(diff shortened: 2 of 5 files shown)


scripts/test.sh
```diff
@@ -1,5 +1,6 @@
 #!/usr/bin/env bash
 set -e
+set -x
 cd "$(dirname "$0")"
```
krzkaczor: are changes in this file really vital for this PR?

quezak: not at all, I just found the test script output a bit unreadable -- I had some errors, but didn't know from which command or test file they originated

krzkaczor: Can we just revert those? I use `-x` only for debugging purposes, I don't leave that in prod code (and this is how all other scripts are written).

PR #138, opened 2019-02-09
Generated type should inherit from web3 Contract

nambrot:
All generated types with the web3 target should inherit from `Contract` if I see correctly. In my use case, I all the contract instances to inherit from `Contract` so that I can take arbitrary contracts for parameter parsing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,14 @@
 # Changelog
 
+## [0.3.13] - 2019-03-14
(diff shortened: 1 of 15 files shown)


nambrot:
@krzkaczor Was wondering if you could take a look at this PR?

nambrot:
Ahh nevermind, saw that it was fixed in https://github.com/ethereum-ts/TypeChain/pull/141

krzkaczor:
Apologize @nambrot — I somehow missed your PR.

PR #144, opened 2019-03-14
Output type is "number" instead of "BigNumber" when bits <= 48

### LINKED ISSUE #142: [ethers] uint8 returned data is infered as BigNumber, but ethers return number instead
For constant methods (at least), uint8 numbers (at least) are returned as numbers by ethers.js Contract, but TypeChain generate bindings as if they were `BigNumber`.

andrevmatos:
Fix #142

### DIFF
--- a/__snapshots__/DumbContract.spec.ts.js
+++ b/__snapshots__/DumbContract.spec.ts.js
@@ -184,6 +184,15 @@ export class DumbContract extends TC.TypeChainContract {
         stateMutability: "nonpayable",
         type: "function",
       },
+      {
+        constant: true,
+        inputs: [],
+        name: "returnSmallUint",
+        outputs: [{ name: "", type: "uint8" }],
+        payable: false,
+        stateMutability: "pure",
+        type: "function",
+      },
       { inputs: [], payable: false, stateMutability: "nonpayable", type: "constructor" },
       {
         anonymous: false,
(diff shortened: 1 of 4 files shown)


krzkaczor:
Thank you for your contribution! 👍 

Could you add a test proving that it works? Take a look at DumbContract.sol. You might want to add there repro case (just a function returning uint8 for example). And then you can write new test case in `test/integration/targets/ethers/DumbContract.spec.ts` Thanks!

krzkaczor:
@andrevmatos i think you need to update snapshots. Try `SNAPSHOT_UPDATE=1 yarn test`

andrevmatos:
Done.

__snapshots__/DumbContract.spec.ts.js
```diff
@@ -230,6 +239,10 @@ export class DumbContract extends TC.TypeChainContract {
     return TC.promisify(this.rawWeb3Contract.byteArray, []);
   }
 
+  public get returnSmallUint(): Promise<BigNumber> {
```
krzkaczor: note: these snapshots are used by different target thats why it's BigNumber here

krzkaczor:
@andrevmatos Thanks! Great work! 🥇

krzkaczor:
Released as `typechain@0.3.13`

PR #145, opened 2019-03-21
correct a typing of web3 generator

### LINKED ISSUE #143: [web3-1.0] Events have incorrect fromBlock type
As of `0.3.12` which uses `1.0.0-beta.48` events have an incorrect `fromType`:

```typescript
events: {
    Transfer(
      options?: {
        filter?: object;
        fromBlock?: Block;
        topics?: (null | string)[];
      },
      cb?: Callback<EventLog>
    ): EventEmitter;
}
```

The `fromBlock` has a type of `Block`, which in web3 looks like: 

```typescript
interface Block extends BlockHeader {
    transactions: Transaction[];
    size: number
    difficulty: number
    totalDifficulty: number
    uncles: string[];
}
```

However it should have a type of `number | string`

cryptokat:
fix #143

### DIFF
--- a/lib/targets/web3/generation.ts
+++ b/lib/targets/web3/generation.ts
@@ -39,7 +39,7 @@ export function codegen(contract: Contract) {
       allEvents: (
           options?: {
               filter?: object;
-              fromBlock?: Block;
+              fromBlock?: number | string;
               topics?: (null|string)[];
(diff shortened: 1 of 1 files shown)


krzkaczor:
Thanks!

Issue #123, opened 2018-12-02
ethers: bytes argument generating the wrong type


pkieltyka:
great project and works quite well with ethers v4

however, for a function declaration like `function safeTransferFrom(address _from, address _to, uint256 _id, uint256 _amount, bytes _data)` typechain for ethers generates the following declaration:

```typescript
    safeTransferFrom(
      _from: string,
      _to: string,
      _id: number | string,
      _amount: number | string,
      _data: (string)[]
    ): Promise<ContractTransaction>;
```

the `_data: (string)[]` argument is wrong, and should be: `_data: Uint8Array`

krzkaczor:
Thanks for the report! I think we are gonna remove ethers target because AFAIK @ricmoo implemented type generator as a part of ethersjs CLI. Can someone confirm it?

pkieltyka:
oh? as far as I now, ethers-cli is a deploy tool, not really a typescript declaration generator from an abi -- but maybe there's a separate tool somewhere

ricmoo:
I haven’t pushed it yet, because it does support tuples yet. Should I push it now and worry about tuples later?

ricmoo:
(maybe after adding the tuple type to human-readable ABI?)

pkieltyka:
@ricmoo no harm is pushing it early and just noting tuples support is coming, and make an issue in there about it..

snario:
+1 would love to try this @ricmoo

dhl:
Would love to see the declaration generator released also, especially if TypeChain is going to phase out support for ethers.js.

ricmoo:
I spent today porting ethers to try using lerna, which I’m liking so far. :)

I’ll try to get my tool released in the `next` branch of ethers-cli tomorrow. :)

snario:
@ricmoo do you have an example of usage?

quezak:
Any update on this?
BTW: the culprit here looks to be in [parsing the evm types](https://github.com/ethereum-ts/TypeChain/blob/d62a31df81aba7554e08a6f7016cab0a2766e94a/lib/parser/typeParser.ts#L82) -- `bytes` actually gets interpreted as `bytes1[]`, hence in ethers output it's mapped to `string[]`. Isn't this interpretation wrong in the first place? IMHO `bytes` should be mapped to `Buffer | Uint8Array` or something.

```
    case "bytes":
return new ArrayType(new BytesType(1));
```

krzkaczor:
[...]

PR #153, opened 2019-05-22
Add DynamicBytesType evm type for bytes type

### LINKED ISSUE #98: Incorrect type for bytes with web3-1.0.0 target
When using the Web3 1.0.0 target, it seems that Typechain is generating incorrect types for bytes values.

The generated type is `(string | number[])[]`, when I think it should be `string | number[]` (no wrapping array). The generated type does not work with Web3.

Happy to try a PR if someone verifies this is indeed a bug 😄

touhonoob:
to fix #98

### DIFF
--- a/lib/targets/ethers/generation.ts
+++ b/lib/targets/ethers/generation.ts
@@ -14,6 +14,7 @@ import {
   AddressType,
   VoidType,
   BytesType,
+  DynamicBytesType,
   BooleanType,
   ArrayType,
   StringType,
@@ -111,6 +112,7 @@ function generateInputType(evmType: EvmType): string {
     case AddressType:
       return "string";
     case BytesType:
+    case DynamicBytesType:
       return "string";
     case ArrayType:
       return `(${generateInputType((evmType as ArrayType).itemType)})[]`;
@@ -137,6 +139,7 @@ function generateOutputType(evmType: EvmType): string {
     case VoidType:
       return "void";
     case BytesType:
+    case DynamicBytesType:
       return "string";
     case ArrayType:
       return `(${generateOutputType((evmType as ArrayType).itemType)})[]`;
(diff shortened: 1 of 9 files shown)


lib/targets/ethers/generation.ts
```diff
@@ -111,6 +112,7 @@ function generateInputType(evmType: EvmType): string {
     case AddressType:
       return "string";
     case BytesType:
+    case DynamicBytesType:
```
quezak: I think the data type for `bytes` should be `string | Buffer | UInt8Array`, or even just `Buffer | UInt8Array`. In many cases, `bytes` is used to send binary data, for which a `string` is not suitable. @krzkaczor what do you think?

See the comments on issues #98 and #123 for more info.
This comment probably applies to all targets, not only ethers.

lib/targets/truffle/generation.ts
```diff
@@ -123,6 +124,7 @@ function generateInputType(evmType: EvmType): string {
     case AddressType:
       return "string | BigNumber";
     case BytesType:
+    case DynamicBytesType:
```
quezak: Commenting separately from my other suggestion (to add binary types to this mapping), because this looks to be wrong. It's fine to use a `BigNumber` for the `BytesType`, since it's mapped from the `bytes1 .. bytes32` Solidity types, which all fit in a single EVM word (so also a single `BigNumber`). On the other hand, `bytes` can have arbitary length, so it's wrong to use a single `BigNumber` for this.

lib/targets/ethers/generation.ts
touhonoob: Thanks for your review @quezak. That makes total sense to me, but this might a bit out of context for issue #98. I'd prefer to open an independent pull request for this change.

krzkaczor: The original issue was about web3 target so i agree that it doesnt make sense to fix it here 🤷‍♂ I will create a new issue when this is done.

touhonoob:
@krzkaczor rebased

krzkaczor:
Merged, thanks for contribution @touhonoob !

PR #156, opened 2019-06-05
Add struct output types for ethers

### LINKED ISSUE #119: Add support for return types that are structs
If I have something like...

```solidity
contract C {
    struct Bar {
        address foo;
        uint256 baz;
    }

    function fun() returns (Bar) { ... }
}
```

Then the typing of `C.functions.fun()` is `Promise<object[]>;`.

e-r-w:
Closes #119 

Added typings & tests for struct return types in ethers.js. Seems like this was simply missed for ethers, looks like all other targets already support struct return types


krzkaczor:
Sorry that merging this took so long. Thanks!

Issue #104, opened 2018-09-27
[Truffle] Provide typings for non-transacting contract function calls


m-chrzan:
Truffle provides a way of calling a contract function in a way that returns the result as if the function call were to be executed against the current blockchain state, but without actually affecting the state. See [here](https://github.com/trufflesuite/truffle/tree/develop/packages/truffle-contract#explicitly-making-a-call-instead-of-a-transaction) for documentation. The Truffle target doesn't currently provide typings for these.

## Expected behavior 1

With the following contract code:
```
uint number;

function setGetNumber(uint n) public returns (uint) {
  number = n;
  return number;
}
```
the following TypeScript should fail to compile:
```
const notANumber = [42]
const greeter = await Greeter.new("", { from: deployer });
await greeter.setGetNumber.call(notANumber)
```
but currently compiles.

## Expected behavior 2

With the following contract code:
```
function getNumber() public pure returns (uint) {
  return 42;
}
```
the following TypeScript should successfully compile:
```
const greeter = await Greeter.new("", { from: deployer });
await greeter.getNumber.call()
```
but currently doesn't.

krzkaczor:
Yeah, I had it somewhere on my todo list. Another thing missing is support for fallback functions but ABI parser is lacking it ATM.

ipatka:
I added and tested this functionality in the above PR

krzkaczor:
I guess, it's done now. If not, please reopen.

PR #157, opened 2019-06-10
add non transacting transaction calls and tests to truffle target

ipatka:
https://github.com/ethereum-ts/TypeChain/issues/104

I added call, sendTransaction and estimateGas methods to generation.ts for the truffle target

## Call
call returns the same output as sending the TX but no state changes are made

### Test
I checked the counter method after calling the simulated TX. 
I had to add another countup method to the smart contract which returned a uint to test the simulated return value.

## sendTransaction
sendTransaction returns the TX hash of the state changing method.

### Test
For this test I checked that the method returns a 66 length string.

## estimateGas
estimateGas returns an integer showing how much gas the simulated TX would use.

### Test
For this test I checked that the estimated gas is over the minimum 20k threshold. Could also just check that it is an int

### DIFF
--- a/lib/targets/truffle/generation.ts
+++ b/lib/targets/truffle/generation.ts
@@ -76,9 +76,20 @@ export interface ${c.name}Instance extends Truffle.ContractInstance {
 
 function generateFunction(fn: ConstantFunctionDeclaration | FunctionDeclaration): string {
   return `
-  ${fn.name}(${generateInputTypes(
+  ${fn.name}: {
+    (${generateInputTypes(
+      fn.inputs,
+    )} txDetails?: Truffle.TransactionDetails): Promise<Truffle.TransactionResponse>;
+  call(${generateInputTypes(
     fn.inputs,
-  )} txDetails?: Truffle.TransactionDetails): Promise<Truffle.TransactionResponse>;
+  )} txDetails?: Truffle.TransactionDetails): Promise<${generateOutputTypes(fn.outputs)}>;
+  sendTransaction(${generateInputTypes(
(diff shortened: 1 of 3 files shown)


lib/targets/truffle/generation.ts
```diff
@@ -76,9 +76,20 @@ export interface ${c.name}Instance extends Truffle.ContractInstance {
 
 function generateFunction(fn: ConstantFunctionDeclaration | FunctionDeclaration): string {
   return `
-  ${fn.name}(${generateInputTypes(
+  ${fn.name}: {
+    (${generateInputTypes(
+      fn.inputs,
+    )} txDetails?: Truffle.TransactionDetails): Promise<Truffle.TransactionResponse>;
+  call(${generateInputTypes(
```
krzkaczor: the original issue mentioned that it `call` should be missing on state changing methods which practically is true but in theory you can still `call` them. Probably `sendTransaction` should be missing on pure and view functions as well 🤷‍♂ What do you think?

krzkaczor:
@ipatka great work already but its missing one thing

lib/targets/truffle/generation.ts
ipatka: I think `call` is valid to simulate a state changing method with call and the expected behavior in the issue is not quite right.

`sendTransaction` is already missing on constant/ view/ pure functions as I didn't modify `generateConstantFunction`

krzkaczor: Ok I agree 👍

krzkaczor:
Thank you very much for contribution 👍

PR #158, opened 2019-06-17
[ethers] named arguments for function args encoding methods

quezak:
The `interface.functions.fnName.encode` methods that I added in PR #154 were effectively typed as, for example, `encode(params: [string, string, string]): string`. Though typesafe, this wasn't really helpful in developement.

With this change, I'm overriding not only the arguments typing for `encode`, but the whole method, and destructuring the arguments so their original names are shown, for example: `encode([receiverAddress, signature, assetData]: [string, string, string]): string`. I believe this is the best we can do here -- because ethers accepts the arguments as an array here, we can't just normally list the arguments with names.

Same problem solved for `encodeTopics` method on `interface.events.eventName`.

No functionality or typing is changed, only the names are added.

### DIFF
--- a/lib/targets/ethers/generation.ts
+++ b/lib/targets/ethers/generation.ts
@@ -93,7 +93,9 @@ function generateEstimateFunction(fn: FunctionDeclaration): string {
 
 function generateInterfaceFunctionDescription(fn: FunctionDeclaration): string {
   return `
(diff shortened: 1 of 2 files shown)


krzkaczor:
Released as `0.3.16`

PR #159, opened 2019-06-18
[ethers] fix argument types for bytes & numbers

### LINKED ISSUE #123: ethers: bytes argument generating the wrong type

quezak:
Closes #123 

I updated the generated typings for ethers, so that number and byte-array arguments accept precisely what do [the docs say](https://docs.ethers.io/ethers.js/html/api-contract.html#types) -- especially passing a `Buffer` for the binary `bytes/bytesN` types.

For a cleaner code I used the `BigNumberish` and `Arrayish` type unions exported by ethers, since that's exactly their job :) It results in better DX than `number | string | Buffer | UInt8Array`, which gets hard to read when you have 4 or more arugments to a contract function.

Updated the tests to check all the variants.

### DIFF
--- a/lib/targets/ethers/generation.ts
+++ b/lib/targets/ethers/generation.ts
@@ -26,7 +26,7 @@ export function codegen(contract: Contract) {
   const template = `
   import { Contract, ContractTransaction, EventFilter, Signer } from 'ethers';
   import { Listener, Provider } from 'ethers/providers';
-  import { BigNumber, Interface } from "ethers/utils";
+  import { Arrayish, BigNumber, BigNumberish, Interface } from "ethers/utils";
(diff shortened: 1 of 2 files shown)


krzkaczor:
Released as `typechain@0.3.17`

PR #160, opened 2019-07-03
[ethers] Generate contract factories

### LINKED ISSUE #137: [ethers] Generate constructor passing ABI to ethers
Hi, 

right now ethers target only generates type definitions, leaving the user with need to manually instantiate the ethers `Contract` class, passing the ABI to it, and then cast the result to a correct type.
So, basically,  right now something like this code is required:
```typescript
import { Contract } from 'ethers';
import { Token } from 'token-contract/types/Token';
import { TokenABI } from 'token-contract/abi/Token.json';
...
const token: Token = new Contract(address, TokenABI, provider) as Token;
```
It would be nice, if `ethers` target generated typescript files having constructor implemented which passed contract ABI to base class (similar how legacy target does it now).
So, new generated code would look something like:
```typescript
const ABI = [
  ...
  <abi>
  ...
];

export class Agreement extends Contract {
  constructor(nameOrAddress: string | Address, providerOrSigner: Provider | Signer) {
    super(nameOrAddress, ABI, providerOrSigner);
  }

  functions: {
     ...
   }
}
```
This would ease the use of generated classes, making only this implementation needed to be exported from package (instead of type definitions and the ABI).

P.S. 
We would love to implement that kind generation in [Waffle](https://getwaffle.io) to work out-of-the-box.

quezak:
Fixes #137

ethers v5 beta doesn't look ready to be used yet, so I opted to solve #137 in TypeChain for the time being.

There are 2 options for this PR -- tell me what do you think.

### 1. Only a type declaration for the contract factories
(see only commit 11f380f)
In this variant, we keep the current behavior for ethers, where TypeChain only generates typings, with no runtime code whatsoever. This adds a simple `<ContractName>Factory` typing declaration, which helps to properly type constructor arguments for contract deployment -- but the ABI and bytecode must still be passed manually (usually re-reading the JSONs on disk).

### 2. Fully working contract factories
(full PR)
In this variant, we deviate a bit from the current behavior: the contract typing files are left as they are now, but TypeChain also generates a separate runtime `<ContractName>Factory` class, which extends ethers' `ContractFactory`, but has the ABI and bytecode built-in. Now, it's super easy to deploy contracts -- just instantiate the factory and pass the constructor arguments, no file operations needed.

We might want to add a CLI switch that enables/disables creation of the concrete factories (when off, it might generate the factory typings from first option instead). Is there a mechanism for "individual generator options", or would this have to be declared next to TypeChain's global CLI options?

### DIFF
--- a/lib/parser/abiParser.ts
+++ b/lib/parser/abiParser.ts
@@ -254,3 +254,33 @@ export function extractAbi(rawJson: string): RawAbiDefinition[] {
 
   throw new MalformedAbiError("Not a valid ABI");
 }
+
+export function extractBytecode(rawContents: string): string | null {
+  const bytecodeRegex = /^[0-9a-fA-F]+$/;
+  // First try to see if this is a .bin file with just the bytecode, otherwise a json
+  if (rawContents.match(bytecodeRegex)) return rawContents;
+
+  let json;
+  try {
+    json = JSON.parse(rawContents);
+  } catch {
+    return null;
+  }
+
+  if (!json) return null;
+
+  if (json.bytecode && json.bytecode.match(bytecodeRegex)) {
+    return json.bytecode;
+  }
+
+  if (
+    json.evm &&
+    json.evm.bytecode &&
+    json.evm.bytecode.object &&
+    json.evm.bytecode.object.match(bytecodeRegex)
+  ) {
+    return json.evm.bytecode.object;
+  }
+
+  return null;
+}
--- a/lib/parser/abiParser.ts
+++ b/lib/parser/abiParser.ts
@@ -256,9 +256,9 @@ export function extractAbi(rawJson: string): RawAbiDefinition[] {
 }
 
 export function extractBytecode(rawContents: string): string | null {
-  const bytecodeRegex = /^[0-9a-fA-F]+$/;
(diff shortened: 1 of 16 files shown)


quezak:
Fixed coverage ;)

quezak:
I was so focused on deployments, that I forgot that the regular ethers contract constructor, that connects to an existing deployed contract, also requires the ABI to be passed in arguments :)

With the second approach (full runtime factory) this can be worked around -- instead of using `new ContractName(address, <abi>, signer)` we could use `new ContractNameFactory(signer).attach(address)`. If we wanted to simplify the contract constructor to be called without the ABI, we'd need to have actual runtime contract classes, not only type declarations for them.

lib/parser/abiParser.ts
```diff
@@ -254,3 +254,38 @@ export function extractAbi(rawJson: string): RawAbiDefinition[] {
 
   throw new MalformedAbiError("Not a valid ABI");
 }
+
+export function extractBytecode(rawContents: string): string | null {
```
krzkaczor: Please use `undefined` instead of `null`. We should prefer undefined in our codebase (generated code is a different thing

krzkaczor:
Published as `typechain@0.3.18` Thanks!

PR #162, opened 2019-07-10
[ethers] Factory for connecting to existing contract instances

quezak:
This is a follow-up for PR #160 (CC issue #137)

Problem: existing methods allowed only to deploy a new contract instance, or attach to an existing address, but only connecting through a complete `Signer` / `Wallet`. In actual use cases, you often don't need (or even _don't want_) the contract instance to have a signer connected, just a provider, to use contract's `view` or `pure` functions.

Proposed solution:
 * add to the contract factories: `static connect(address: string, signerOrProvider: Signer | Provider)` -- this just invokes the `Contract` constructor with the same params and the hardcoded ABI (since we can't add runtime functions to our contract type declarations)
 * for abstract contracts (or if bytecode was just not provided), generate a simplified version of the factory, containing just the above static method, without the usual deployment methods.

### DIFF
--- a/__snapshots__/DumbContract.spec.ts.js
+++ b/__snapshots__/DumbContract.spec.ts.js
@@ -176,6 +176,15 @@ export class DumbContract extends TC.TypeChainContract {
         type: "function",
       },
       {
+        constant: true,
+        inputs: [],
+        name: "testVoidReturn",
+        outputs: [],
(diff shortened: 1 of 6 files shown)


krzkaczor:
Published as `typechain@0.3.19`

Issue #106, opened 2018-09-28
ABI Parser extraction and improvements


krzkaczor:
ABI Parser should be extracted to separate library and improved:
- return discriminative unions
- add support for lacking types:
	- fallback function 
    - new `stateMutability` option (https://github.com/ethereum-ts/TypeChain/issues/65)
	- overloading methods and constructors
    - tuple/struct types (https://github.com/ethereum-ts/TypeChain/issues/105)

Issue #65, opened 2018-07-26
Support stateMutability


axic:
For close to a year now Solidity also outputs the `stateMutability` field for each function. This can have the following values:
- pure
- view
- nonpayable
- payable

For backwards compatibility also the field `constant` (set to true on `stateMutability` pure or view) and `payable` (set to true on `stateMutability `payable`) is present.

We'd like to phase out the `constant` and `payable` fields. For more see https://solidity.readthedocs.io/en/develop/abi-spec.html#json

It seems that `abiParser.ts` doesn't support it: https://github.com/Neufund/TypeChain/blob/master/lib/abiParser.ts#L47

krzkaczor:
@axic thanks for your comment. I actually really like this new approach. I will work on improving it when i ll find some time.

Issue #150, opened 2019-04-08
Support Function Overloading


PhABC:
As of now, it seems like generated types do not support function overload. Would it be possible to add support for this?

This feature has been mentioned in #106 but I haven't seen any update since that issue was opened.

krzkaczor:
It's supported on ABI parser level. Now support for all targets is missing.

pash7ka:
@krzkaczor I can work on this, if you give me some clue how to approach this problem. As i understand, there is no support in TypeScript for overloading functions with different signatures, it only allows to ommit some parameters.
My current work-around is to convert to `<any>` and then use `methods` field of a contract oblect:
```
let instance = await MyInitializableContract.new();
await (<any> instance ).methods['initialize(address)'](myInitData, {from: owner});
```
One approach would be to have functions with names like `method_0xc4d66de8` where `0xc4d66de8` is first 4 bytes of `keccak256('initialize(address)')`.

krzkaczor:
@pash7ka I am glad that you're up for the task ;) So let me start by emphasizing that we don't come up with our own API but rather conform to API that is used by target library. 

As far as I remember most of libs choose somehow default (first in abi?)  implementation (so for example named simply `initialize`) and then use full names (with args) for overloads (for example: `initialize(address)`). So here we don't actually need any overloading on the level of generated wrappers - there are simply functions with different names.

TypeChain will group entities (methods/events) by name so it's quite easy to generate code for all overloads, for now we just pick first impl.

Let me know if you have any questions, for starters just focus on one target. Probably in few weeks (or even sooner) I will work on this by myself because it's a high time 😆

pash7ka:
>As far as I remember most of libs choose somehow default (first in abi?) implementation 

For example, Web3.js implementaions has special handling for such case. They [create a linked list of implementations](https://github.com/ethereum/web3.js/blob/1.x/packages/web3-eth-contract/src/index.js#L134) and [iterate through it](https://github.com/ethereum/web3.js/blob/1.x/packages/web3-eth-contract/src/index.js#L774) until number of arguments does not match.

> then use full names (with args) for overloads (for example: `initialize(address)`). 

I believe TypeScript does not allow parenthesis in function names. 

Let's get to example. Currently, for truffle, i may have:
```
export interface IMyInitializableContractInstance extends Truffle.ContractInstance {
  initialize: {
[...]

PR #172, opened 2019-09-21
TypeChain ver 1.0.0

krzkaczor:
## Goals:
- [x] complete ABI parsing to typesafe structures
  - [x] use a discriminative union for EvmTypes
  - [x] support overrides
- [x] split targets to separate packages
- [x] support custom targets
- [x] add prebublish hooks for subpackages
- [x] better error messages on missing target packages
- [ ] readme
- [ ] readme for each package
- [x] revive truffle example

Related https://github.com/ethereum-ts/TypeChain/issues/150
Closes https://github.com/ethereum-ts/TypeChain/issues/65
Closes https://github.com/ethereum-ts/TypeChain/issues/106


packages/core/lib/parser/parseEvmType.ts
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  if (isIntTypeRegex.test(rawType)) {
+    const match = isIntTypeRegex.exec(rawType);
+    return { type: "integer", bits: parseInt(match![1] || "256") };
+  }
+
+  if (isBytesTypeRegex.test(rawType)) {
+    const match = isBytesTypeRegex.exec(rawType);
+    return { type: "bytes", size: parseInt(match![1] || "1") };
```
quezak: The regex is `/^bytes([0-9]+)$/` with a `+`, so `match![1]` can't be empty (and it can't be 0, since solidity wouldn't compile that)

packages/typechain-target-ethers/lib/generation.ts
```diff
@@ -242,54 +245,51 @@ function generateEventArgType(eventArg: EventArgDeclaration): string {
...
+    case "string":
       return "string";
-    case TupleType:
-      return generateTupleType(evmType as TupleType, generateInputType);
-
+    case "tuple":
+      return generateTupleType(evmType, generateInputType);
     default:
-      throw new Error(`Unrecognized type ${evmType.constructor.name}`);
+      throw new UnreachableCaseError(evmType);
```
quezak: is the `default` case even necessary? now that we have an exhaustive switch over a discriminated union, it woudl be even better to remove the `default` case, so the compilation fails if we forget a case above

quezak: Same question for output type switch below, and for other targets.

quezak:
I'll try to test the new version with our project (on ethers) later today, to see if there are no regressions.

quezak:
Just tested the newest beta version with ethers target, works fine!

Two issues though:

1. At first, I didn't install the `typechain-target-ethers` package, and the error message is rather cryptic. If I didn't read the code before, I would have no clue I have to install it. If target is one of the "official" targets, the error should say to install the package, and for other targets, say something about invalid target package (or a link to custom targets FAQ in the future).
```
   0[quezak@jagdsabel]~/...>: npx typechain --target ethers --outDir $TYPECHAIN_OUT_DIR "build/contracts/*.json"
Error occured:  module.default is not a constructor
```

2. I've reviewed changes in the generated files. The big difference is that the functions are now also available on contract object's toplevel, which is correct with how ethers generates them. The difference I'm curious about is that methods for contract variable getters -- they now also have the `transactionOverrides` parameter. It's not necessary there, maybe even invalid -- constant functions are not called as transactions. Example:

   * solidity:
```
    bool public isActive = true;
    address public owner;
```

   * typechain 0.3.19:
```
    isActive(): Promise<boolean>;
    owner(): Promise<string>;
```

   * typechain 1.0.0-beta.2 + typechain-target-ethers 1.0.0-beta.1:
```
    isActive(overrides?: TransactionOverrides): Promise<boolean>;
    owner(overrides?: TransactionOverrides): Promise<string>;
```

packages/core/lib/parser/parseEvmType.ts
```diff
@@ -0,0 +1,93 @@
...
+const isIntTypeRegex = /^int([0-9]*)$/;
+const isBytesTypeRegex = /^bytes([0-9]+)$/;
+
+export function parseEvmType(rawType: string, components?: EvmSymbol[]): EvmType {
+  const lastChar = rawType[rawType.length - 1];
+
+  // first we parse array type
+  if (lastChar === "]") {
+    let finishArrayTypeIndex = rawType.length - 2;
+    while (rawType[finishArrayTypeIndex] !== "[") {
```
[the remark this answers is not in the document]
krzkaczor: this was just copied over. refactoring is possible in the future but first I would like to have better test coverage

packages/core/lib/parser/parseEvmType.ts
```diff
@@ -0,0 +1,93 @@
...
+      finishArrayTypeIndex--;
+    }
+
+    const arraySizeRaw = rawType.slice(finishArrayTypeIndex + 1, rawType.length - 1);
+    const arraySize = arraySizeRaw !== "" ? parseInt(arraySizeRaw) : undefined;
+
+    const restOfTheType = rawType.slice(0, finishArrayTypeIndex);
```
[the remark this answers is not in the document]
krzkaczor: same

packages/core/lib/parser/abiParser.ts
```diff
@@ -289,3 +280,18 @@ export function ensure0xPrefix(hexString: string): string {
...
+    (fn.stateMutability === "pure" || fn.stateMutability === "view") &&
+    fn.inputs.length === 0 &&
+    fn.outputs.length === 1
+  );
+}
+
+export function isConstantFn(fn: FunctionDeclaration): boolean {
+  return (
+    ((fn.stateMutability === "pure" || fn.stateMutability === "view") && fn.inputs.length !== 0) ||
+    fn.outputs.length !== 1
```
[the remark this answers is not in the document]
krzkaczor: I think you're right. In fact, I fixed it already once 😆 i need to write tests for this part as its trying to behave in the same way as code from previous version.

packages/typechain-target-ethers/lib/generation.ts
krzkaczor: Oh, actually it's not needed with explcit signature like this 👍 good point

krzkaczor:
ad 2. it can be related to a bug in `isConstantFn`. Thanks for bringing this up!

krzkaczor:
I just fixed ad2 👍, I updated todo in the PR description.

krzkaczor:
ad 1, oh actually the error message was supposed to be much more civilized but you managed to hit another bug 😆 Fixing it now. It will say something like:
```
$ npx typechain --target x
Error occured:  Couldnt find x. Tried loading: typechain-target-x, /Users/krzkaczor/Workspace/ethereum-ts/TypeChain/x.
Perhaps you forgot to install typechain-target-x?
```

eswarasai:
@krzkaczor - I got the below error message when I haven't installed `web3-1.0.0` target.

```
➜  ETHSurvey git:(develop) ✗ npm run typechain 

> ethsurvey@0.0.1 typechain ~/ETHSurvey
> typechain --target=web3-1.0.0 --outDir ./src/types './build/contracts/*.json'

Error occured:  Couldnt find web3-1.0.0. Tried loading: typechain-target-web3-1.0.0, ~/ETHSurvey/web3-1.0.0.
Perhaps you forgot to install typechain-target-web3-1.0.0?
```

When I tried to do `npm install typechain-target-web3-1.0.0`, it threw a missing package error. As the package available on npm is under name `typechain-target-web3-1`. We need to fix the error message being displayed as well as probably change the targets now to `web3-v1` and `web3-v2` rather than `1.0.0` and `2.0.0`. Let me know your thoughts. Thanks!

krzkaczor:
Hey!
So the name of the target changed from `web3-1.0.0` to `web3-1`. TBH I don't see anything particularly wrong with this naming scheme ie. `web3-1` and `web3-2` but maybe adding `v` would make it clearer. @quezak thoughts?

quezak:
yeah, I think `web3-v1` would be clearer. `web3-1` souds like it's a `web` package with version `3.1` :D `web3-1.0.0` is also confusing, as it indeed looks like a version specification for web3, and not related to typechain.

packages/typechain-target-ethers/README.md
```diff
@@ -0,0 +1,19 @@
...
+    <a href="https://coveralls.io/github/ethereum-ts/TypeChain?branch=master"><img alt="Coverage" src="https://coveralls.io/repos/github/ethereum-ts/TypeChain/badge.svg?branch=master"></a>
+    <img alt="Downloads" src="https://img.shields.io/npm/dm/typechain.svg">
+    <a href="https://github.com/prettier/prettier"><img alt="Prettier" src="https://img.shields.io/badge/code_style-prettier-ff69b4.svg"></a>
+    <a href="/package.json"><img alt="Software License" src="https://img.shields.io/badge/license-MIT-brightgreen.svg?style=flat-square"></a>
+  </p>
+
+  <p align="center">
+    <a href="https://blog.neufund.org/introducing-typechain-typescript-bindings-for-ethereum-smart-contracts-839fc2becf22">Medium post</a> | <a href="https://www.youtube.com/watch?v=9x6AkShGkwU">DappCon Video</a>
+  </p>
+</p>
```
quezak: We could add a link to the main TypeChain README to all target package readmes -- in case someone lands there from a google search or something.

eswarasai:
@krzkaczor - Let me know if you need any help here in wrapping up this PR. I'll get started with the `web3-v2` target once this is merged 🙂

quezak:
@krzkaczor heads up -- I've renamed the `web3-v1` integration test dir too, plus some other occurences you missed.

PR #176, opened 2019-10-01
Fix failure case when fallback function is provided

HenryNguyen5:


packages/core/lib/parser/abiParser.spec.ts
```diff
@@ -101,3 +103,17 @@ describe("parseEvent", () => {
...
 });
+
+describe("parse", () => {
+  describe("fallback functions", () => {
+    it("should work on output-less fallback functions", () => {
+      const fallbackAbiFunc: RawAbiDefinition = {
+        payable: true,
+        stateMutability: "payable",
+        type: "fallback",
+      } as any;
```
HenryNguyen5: Had to mark this as `any` because it doesnt conform to `RawAbiDefinition`. The interface needs to have more properties marked as optional but it was out of scope for the fix we need

krzkaczor:
Released as `typechain@1.0.0-beta.2`

Issue #125, opened 2018-12-10
Unknown type: TokenLib.ModuleData storage


shuffledex:
Trying to compile this part of json:

```
    {
      "constant": false,
      "inputs": [
        {
          "name": "_moduleData",
          "type": "TokenLib.ModuleData storage"
        },
        {
          "name": "_module",
          "type": "address"
        }
      ],
      "name": "archiveModule",
      "outputs": [],
      "payable": false,
      "stateMutability": "nonpayable",
      "type": "function"
    }
```

I'm getting `Unknown type: TokenLib.ModuleData storage` and the execution stoped.

krzkaczor:
@shuffledex Frankly, I have no idea what `TokenLib.ModuleData storage`is. Do you have source code of the smart contract that generated such ABI?

stwiname:
It will be referring to a [Library](https://solidity.readthedocs.io/en/latest/contracts.html#libraries)

Example in link includes `Data storage data`

quezak:
@shuffledex can you provide a full minimal example contract and library that generate the json you posted? I'll probably soon be working on linking contracts with (non-external) libraries, so I may take a look at your case too.

krzkaczor:
This should be fixed in `typechain@1.0.2`, if not please reopen and provide repro ;)

stwiname:
I'm still facing this issue after updating to `typechain@1.0.3`. I've attached a zip of the JSON files generated from `truffle@5.0.32` compiler.

Let me know if I can be more help. I can create a more complete example if necessary.

[Archive.zip](https://github.com/ethereum-ts/TypeChain/files/3766439/Archive.zip)

quezak:
OK, leaving this issue closed, continuing in #216

PR #178, opened 2019-10-10
[ethers] link internal libraries in factory

### LINKED ISSUE #177: [ethers] link internal libraries in factory before deployment
Currently, contract using libraries that were not provided in `solc` parameters, are treated as abstract -- which is kinda correct, but for internal libraries, compiled together with the contracts, we want to be able to deploy them first, and then link them to the contract prior to their deployment.

Todo:
* [x] next to `extractBytecode`, add `extractBytecodeAndLinkReferences`, which matches the bytecode with library address placeholders and also returns the `linkReferences` object
* [x] for contracts which don't have a bytecode, but do have a bytecode with link references, generate a different factory, that also takes a dictionary of lib addresses in parameters
* [x] see if the solution is the same/similar for other targets than ethers

Example json of a contract that uses an API:
```
    "bytecode": {
      "linkReferences": {
        "contracts/testWithLib.sol": {
          "TestLib": [
            {
              "length": 20,
              "start": 151
            }
          ]
        }
      },
      "object": "....0e11b815273__$cf80b356982734390a54f3eb79e60a0c87$__9163a892066e9161009e9190600401610163565b600....",
```

Example explanation:
For each call to a library function, the compiler puts an entry in `linkReferences: { <libPath>: { <libName>: ... }}`, and instead of the address of the contract to call, outputs a link placeholder.
The placeholder is `__$<libHash>$__`, where ``libHash = solidityKeccak256(['string'], `${libPath}:${libName}`).slice(2, 36)``.

quezak:
* closes #177 
* the parser also extracts references to unlinked libraries from the bytecode.
* if the contract json contains the `evm.bytecode.linkReferences` (which is the case e.g. when using [waffle](https://github.com/EthWorks/Waffle)), its contents are used to get human understandable keys for the library addresses object (ie. `contracts/file.sol:SomeContract` instead of a partial hash).
* the generated contract factory for `ethers` target will have another param to specify addresses of the libraries to be linked.
* other targets don't generate factories for now, but if they do, they will be able to use the updated parser `extractBytecode` function to easily implement similar linking feature.
* NOTE: the base for this PR is `kk/parser-rewrite` for easier review. It should be changed to `master` once #172 is merged.

EDIT by krzkaczor:
Closes https://github.com/ethereum-ts/TypeChain/issues/125

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,5 @@ test/integration/targets/web3-v1/types/web3-contracts
 test/integration/targets/ethers/types/ethers-contracts
 test/package-test/types
 packages/core/README.md
+.vscode
+*.code-workspace
--- a/.gitignore
+++ b/.gitignore
@@ -11,5 +11,3 @@ test/integration/targets/web3-v1/types/web3-contracts
 test/integration/targets/ethers/types/ethers-contracts
 test/package-test/types
 packages/core/README.md
-.vscode
-*.code-workspace
--- a/packages/core/lib/parser/abiParser.spec.ts
+++ b/packages/core/lib/parser/abiParser.spec.ts
@@ -1,19 +1,19 @@
 import { expect } from "chai";
+import { merge } from "lodash";
 
 import { MalformedAbiError } from "../utils/errors";
 import {
   ensure0xPrefix,
   extractAbi,
   extractBytecode,
-  parseEvent,
-  RawEventAbiDefinition,
-  parse,
-  RawAbiDefinition,
   FunctionDeclaration,
   isConstant,
   isConstantFn,
+  parse,
+  parseEvent,
+  RawAbiDefinition,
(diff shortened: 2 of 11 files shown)


quezak:
(^ the force-push above was just a rebase to the updated base branch)

.gitignore
```diff
@@ -11,3 +11,5 @@ test/integration/targets/web3-v1/types/web3-contracts
 test/integration/targets/ethers/types/ethers-contracts
 test/package-test/types
 packages/core/README.md
+.vscode
```
krzkaczor: use your global gitignore file please :) otherwise this file will become global list of all IDE's cache directories

krzkaczor:
Released as:
- `typechain@1.0.2`
- `typechain-target-ethers@1.0.1`

PR #179, opened 2019-10-18
Handle 0x/sol-compiler output

HenryNguyen5:


krzkaczor:
Thanks! Released as `typechain@1.0.1`.

PR #181, opened 2019-10-21
Handle @0x/sol-compiler bytecode output

HenryNguyen5:


packages/core/lib/parser/abiParser.ts
```diff
@@ -292,6 +292,20 @@ export function extractBytecode(rawContents: string): BytecodeWithLinkReferences
     return extractLinkReferences(json.evm.bytecode.object, json.evm.bytecode.linkReferences);
   }
 
+  // handle json schema of @0x/sol-compiler
+  if (
+    json.compilerOutput &&
+    json.compilerOutput.evm &&
+    json.compilerOutput.evm.bytecode &&
+    json.compilerOutput.evm.bytecode.object &&
```
krzkaczor: i think it would be much more readable to use lodash's `get` here

krzkaczor: (or optional chaining once TS 3.7 is released :->)

HenryNguyen5: Yeah, cant wait for optional chaining. I didnt bother with `get` because the code block above uses the same structure to check for nested objects. Just wanted to keep it consistent. Do you want me to change it still @krzkaczor?

krzkaczor: ok makes sense. I am gonna rewrite it to optional chainging soon 😈

krzkaczor:
Released as `typechain@1.0.3`

PR #184, opened 2019-11-06
Switches BigNumber to BN.

### LINKED ISSUE #136: Wrong big number instance type
Generated types references bignumber.js but in reality bn.js is used by ethereum.

MicahZoltu:
Truffle (at least version 5.0.x) use `BN` internally, not BigNumber.

Fixes #136.

### DIFF
--- a/packages/typechain-target-truffle/lib/generation.ts
+++ b/packages/typechain-target-truffle/lib/generation.ts
@@ -14,7 +14,7 @@ import { values } from "lodash";
 export function codegen(contracts: Contract[]) {
   const template = `
 /// <reference types="truffle-typings" />
-import { BigNumber } from "bignumber.js";
+import BN from "bn.js";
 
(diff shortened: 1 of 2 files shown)


krzkaczor:
Hey! First of all thanks for your contribution! 

I think that the root of the problem is that our current truffle target works with `truffle@4`. `truffle@5` uses web3.js under the hood so for sure there are breaking changes.

What I think we need to do is:
- rename current target as `truffle@4`
- create new target for `truffle@5`

So, I think this is unmergable right now :( 

This is a bigger task, let me know @MicahZoltu if you would be interested in creating new target for `@5` and properly fixing this problem. I can request gitcoin bounty for this.

MicahZoltu:
No, I'm not interested in taking on a bigger task.  😄  I thought this task was perhaps quick and easy and just needed someone to throw together a PR, but if it requires more than the 5 minutes I put into this then I'll leave it for someone else.

PR #186, opened 2019-11-15
Handle absolute paths in outDir

spalladino:
The previous implementation joined the user-provided outDir with the working
directory. This meant that an absolute path as an outDir would get concatenated
to the working dir. Using path.resolve instead handles both absolute and
relative paths properly.

```
> resolve(process.cwd(), '/foo')
'/foo'
> resolve(process.cwd(), 'foo')
'/home/spalladino/Projects/TypeChain/foo'
> resolve(process.cwd(), './foo')
'/home/spalladino/Projects/TypeChain/foo'
```

### DIFF
--- a/packages/typechain-target-ethers/lib/index.ts
+++ b/packages/typechain-target-ethers/lib/index.ts
@@ -1,4 +1,4 @@
-import { join } from "path";
+import { join, resolve } from "path";
 import { Dictionary } from "ts-essentials";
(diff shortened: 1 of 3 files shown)


krzkaczor:
Thanks! Looks good 👍

krzkaczor:
Released as:
`typechain-target-ethers@1.0.2`
`typechain-target-truffle@1.0.1`
`typechain-target-web3-v1@1.0.1`

PR #190, opened 2019-11-20
typechain-target-web3-v1: fixed peerDependency warning appearing during `yarn install`

saniales:
Now
`warning " > typechain-target-web3-v1@1.0.1" has incorrect peer dependency "typechain@1.0.0^".`
Will not appar anymore, as the peerDependency is now specifying correct version


krzkaczor:
I knew that something was wrong with this semver 🤦‍♂ Thanks for tracking this. I ll release new packages in a sec.

Issue #193, opened 2019-11-27
Truffle target generates types not fully compatible with Truffle v5 because BigNumber library was changed to BN.js


pash7ka:
With v5.0.0 Truffle [moved](https://github.com/trufflesuite/truffle/releases/tag/v5.0.0) from [Web3.js](https://github.com/ethereum/web3.js/) v0.x to v1.x. One of major change in it: use of [BN.js](https://github.com/indutny/bn.js/) instead of [bignumber.js](https://mikemcl.github.io/bignumber.js/). This change affects types generated by TypeChain.

This is addressed in a pull-request  https://github.com/ethereum-ts/TypeChain/pull/184, but as @krzkaczor commented there, a new target should be introduced for that.
I've tried to implement this change:  https://github.com/pash7ka/TypeChain/tree/bignumber-to-bn and almost done it, but it required not only cloning target directories, but also renaming packages.
(`typechain-target-truffle` become `typechain-target-truffle-v4`, a new one become `typechain-target-truffle-v5`). 

Should i go forward with this change, which will require all current users of `typechain-target-truffle` to update package name, or it doesn't worth it?
Or maybe just introduce `typechain-target-truffle-v5` and leave `typechain-target-truffle` for v4? (current version of my repository implements this aproach)

krzkaczor:
@pash7ka sorry for the late response. Yes I think you can go ahead without changing anything in the already existing target. We can make the final decision after merging your new target. 

I don't consider renaming an already existing target to be a big deal since it's not breaking anything for already existing users. They just need to remove old package and add new if they want to keep receiving updates.

krzkaczor:
This should be fixed in webv2. Also related to: https://github.com/ethereum-ts/TypeChain/issues/211

https://github.com/ethereum-ts/TypeChain/releases/tag/v2.0.0

PR #196, opened 2019-12-15
Add truffle v5 target, switch from BigNumber to BN

pash7ka:
As mentioned in issue #193 and pull requst #184, with v5.0.0 truffle [moved](https://github.com/trufflesuite/truffle/releases/tag/v5.0.0) from [Web3.js](https://github.com/ethereum/web3.js/) v0.x to v1.x. One of major change in it: use of [BN.js](https://github.com/indutny/bn.js/) instead of [bignumber.js](https://mikemcl.github.io/bignumber.js/).
With this pull request i've introduced `typechain-target-truffle-v5` based on @MicahZoltu pr #184.


packages/typechain-target-truffle-v5/README.md
```diff
@@ -0,0 +1,21 @@
+# Typechain target Truffle
```
krzkaczor: can u add here `Truffle@5` or something?

pash7ka: done

pash7ka:
> Isn't truffle@5 using web3@1 under the hood? Does it mean it has the same api? Can we somehow reuse the generated types between packages?
Yes, they use web3@1 under the hood, butthey created their own wrappers around it.
For example, when [deploying a contract](https://github.com/trufflesuite/truffle/blob/v5.1.4/packages/contract/lib/execute.js#L285) they wrap it with [TruffleContract](https://github.com/trufflesuite/truffle/blob/v5.1.4/packages/contract/lib/contract/constructorMethods.js). Also they [wrap receipt](https://github.com/trufflesuite/truffle/blob/v5.1.4/packages/contract/lib/handlers.js#L132) of a call.

krzkaczor:
@pash7ka oh, makes sense. I will finish CR today 👍

krzkaczor:
Merged to tmp branch for now - it needs a little bit more work. 

Thanks for this @pash7ka @MicahZoltu ! 🎉

PR #218, opened 2020-02-21
[ethers] Add optional deploy tx overrides in contract factories

### LINKED ISSUE #215: ContractFactory violates Liskov substitution principle
Suppose i have a simple contract
```
pragma solidity ^0.6.0;

contract MyContract {
    constructor (address a) public {
    }
}
```

Using typechain i obtain `MyContractFactory`.

Unfortunately `MyContractFactory` lacks functionality compared to the ethers `ContractFactory`.
1. new MyContractFactory(wallet).deploy(...) does not accept overrides as the last argument
2. casting new MyContractFactory(wallet) as ContractFactory and then calling deploy with an incorrect number of arguments we do not get the proper error about argument count from ethers.

Suggested fix (not tested):

```ts
class MyContractFactory {
  deploy(a: string, overrides: providers.TransactionRequest): Promise<MyContract>
  deploy(...args: any[]): Promise<MyContract> {
    super.deploy(...args) as Promise<MyContract>
  }
}
```

quezak:
fixes #215

### DIFF
--- a/test/integration/targets/ethers/types/ethers-contracts/DumbContract.d.ts
+++ b/test/integration/targets/ethers/types/ethers-contracts/DumbContract.d.ts
@@ -36,7 +36,7 @@ interface DumbContractInterface extends Interface {
       encode([boolArrayParam]: [(boolean)[]]): string;
     }>;
 
-    counterArray: TypedFunctionDescription<{ encode([,]: [BigNumberish]): string }>;
+    counterArray: TypedFunctionDescription<{ encode([]: [BigNumberish]): string }>;
 
     countup: TypedFunctionDescription<{ encode([offset]: [BigNumberish]): string }>;
 
--- a/test/integration/targets/ethers/types/ethers-contracts/DumbContract.d.ts
+++ b/test/integration/targets/ethers/types/ethers-contracts/DumbContract.d.ts
@@ -36,7 +36,7 @@ interface DumbContractInterface extends Interface {
       encode([boolArrayParam]: [(boolean)[]]): string;
     }>;
 
-    counterArray: TypedFunctionDescription<{ encode([]: [BigNumberish]): string }>;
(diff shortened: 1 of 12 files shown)


quezak:
@krzkaczor  IDK why the files under ` test/integration/targets/truffle/@types/truffle-contracts/` are updated, they changed after running `yarn test`. Is it just that they should have been changed in one of the earlier PRs but changes were not committed?

test/integration/targets/ethers/types/ethers-contracts/DumbContract.d.ts
```diff
@@ -36,7 +36,7 @@ interface DumbContractInterface extends Interface {
       encode([boolArrayParam]: [(boolean)[]]): string;
     }>;
 
-    counterArray: TypedFunctionDescription<{ encode([]: [BigNumberish]): string }>;
+    counterArray: TypedFunctionDescription<{ encode([,]: [BigNumberish]): string }>;
```
quezak: this `[,]` looks invalid :joy:

sz-piotr:
This looks good, I am however surprised that there is no test file for multiple constructor arguments.

quezak:
> This looks good, I am however surprised that there is no test file for multiple constructor arguments.

Yeah, the test contracts don't have such a constructor. It would be good to add another param to `DumbContract`'s constructor, but let's do that in a separate PR since it will affect tests of all targets and isn't connected to the issue solved here.

krzkaczor:
@quezak it looks like you commited some changes for truffle that occurred only for you... I have no problem re-running tests, build and removing those changes. 

It could be that our (super complicated, convoluted) test pipeline is not deterministic. I will keep this in mind while redesigning testing pipeline this week.

PS: sorry that it took me a full month to take a look 😆

krzkaczor:
Thanks for this!

📦 Released as `typechain-target-ethers@1.0.4`

Issue #214, opened 2020-02-12
Allow shell glob


AuHau:
Hey there,

first of all thanks for a great tool! This really makes much more comfortable to develop with contracts in TS!

One small issue I had is that your CLI args parsing does not allow using shell globbing patterns (eq. shell expansion) because you expect only one glob argument and therefore when shell expands the glob into list of paths the CLI detects only one file instead of all of them. 

Wrapping the glob parameter in parenthesis fix this, but takes time to realize what is going on 😅

krzkaczor:
Yeah, it's PITA, it's even mentioned in readme docs but you're right - this could be supported on CLI level. Thanks for this suggestion 👍

AuHau:
Oh damn, sorry about that 🙈I have missed that in docs. I will have a look if it could be easily worked on and create PR if so.

krzkaczor:
@AuHau you're more than welcome to do so :) 

I think this can be a little bit more complicated because glob handling could be a part of https://github.com/krzkaczor/ts-generator 

anyway, one more reason to merge TypeChain with ts-generator (since it failed to succeed as a separate project) and simplify things.

AuHau:
I see, well I am already too much procrastinating today, so not sure when exactly I will get to it 😅

PR #222, opened 2020-04-20
NatSpec parsing

### LINKED ISSUE #221: NatSpec parsing
I would like to create a documentation from the TypeScript code I am generating. Is parsing and copying the [NatSpec](https://solidity.readthedocs.io/en/v0.5.8/natspec-format.html) comments from the abi (rather the AST) a desired functionality of this library? I would add a PR

chmanie:
This PR is adding functionality to parse and embed NatSpec comments into the generated TypeScript types. It will be using a methodology akin to https://github.com/JoinColony/natspec-typedoc (or use the module itself) to map NatSpec comments to TypeDoc compatible comments. For this to be possible we not only need to extract the ABI from the contracts but also the AST (if available - if not we just won't generate TypeDoc comments). The typechain main package has to be modified to do that. NatSpec is parsed in that module and passed in as another property to the targets which in turn can make use of it.

## TODO

- [x] Extract per-function NatSpec comments from the ast
- [x] Transform NatSpec comments to TypeDoc
- [x] Add `documentation` to the `abiParser.parse` function
- [x] Adjust ethers v4 target
- [x] Adjust truffle v4 target
- [x] Adjust truffle v5 target
- [x] Adjust web3 v1 target
- [ ] Adjust web3 v2 target?

Closes #221.

### DIFF
--- a/packages/target-ethers-v4/src/generation.ts
+++ b/packages/target-ethers-v4/src/generation.ts
@@ -4,6 +4,7 @@ import {
   AbiParameter,
   BytecodeWithLinkReferences,
   Contract,
+  FunctionDocumentation,
   EventArgDeclaration,
   EventDeclaration,
   EvmOutputType,
@@ -200,6 +201,7 @@ function generateLibraryAddressesInterface(contract: Contract, bytecode: Bytecod
 
 function generateFunction(fn: FunctionDeclaration): string {
   return `
+  ${generateFunctionDocumentation(fn.documentation)}
   ${fn.name}(${generateInputTypes(fn.inputs)}${
     !isConstant(fn) && !isConstantFn(fn) ? 'overrides?: TransactionOverrides' : ''
   }): Promise<${
@@ -222,6 +224,27 @@ function generateInterfaceFunctionDescription(fn: FunctionDeclaration): string {
     fn.inputs,
   )}): string; }>;
 `
+}
+
+function generateFunctionDocumentation(doc?: FunctionDocumentation): string {
+  if (!doc) return ''
+
+  let docString = '/**'
+  if (doc.details) docString += `\n * ${doc.details}`
+  if (doc.notice) docString += `\n * ${doc.notice}`
+
+  const params = Object.entries(doc.params || {})
+  if (params.length) {
+    params.forEach(([key, value]) => {
+      docString += `\n * @param ${key} ${value}`
+    })
+  }
+
+  if (doc.return) docString += `\n * @returns ${doc.return}`
+
+  docString += '\n */'
+
+  return docString
 }
 
 function generateInputTypes(input: Array<AbiParameter>): string {
--- a/packages/target-ethers-v4/src/index.ts
+++ b/packages/target-ethers-v4/src/index.ts
@@ -7,6 +7,7 @@ import {
   Contract,
   extractAbi,
   extractBytecode,
+  extractDocumentation,
   getFileExtension,
   getFilename,
   parse,
@@ -79,8 +80,9 @@ export default class Ethers extends TsGeneratorPlugin {
     if (abi.length === 0) {
       return
     }
+    const documentation = extractDocumentation(file.contents)
 
-    const contract = parse(abi, name)
+    const contract = parse(abi, name, documentation)
     const bytecode = extractBytecode(file.contents) || this.bytecodeCache[name]
 
     if (bytecode) {
--- a/packages/target-truffle-v4/src/codegen/contracts.ts
+++ b/packages/target-truffle-v4/src/codegen/contracts.ts
@@ -1,4 +1,4 @@
-import { Contract, FunctionDeclaration, isConstant, isConstantFn } from 'typechain'
+import { Contract, FunctionDeclaration, FunctionDocumentation, isConstant, isConstantFn } from 'typechain'
 import { values } from 'lodash'
 import { codegenInputTypes, codegenOutputTypes } from './types'
 import { codegenEventsDeclarations, codegenAllPossibleEvents } from './events'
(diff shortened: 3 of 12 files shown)


chmanie:
Hey @krzkaczor! I added comment generation for the ethers module. It seems to work fine, I also tested it with our contracts (here's an example: https://github.com/JoinColony/colonyJS/blob/v2/docs/interfaces/_clients_colony_colonyclientv4_.extendedicolonyv4.md#adddomain).

Let me know if this is all good, then I will proceed with the other targets.

packages/target-ethers-v4/src/generation.ts
```diff
@@ -224,6 +226,27 @@ function generateInterfaceFunctionDescription(fn: FunctionDeclaration): string {
 `
 }
 
+function generateFunctionDocumentation(doc?: FunctionDocumentation): string {
```
krzkaczor, line 229: Here we will generate solidity types (like `uint256` for `@returns` and `@param`) for people using typescript... I am not sure if this won't confuse IDEs (typescript type annotation will clash).

What do you think about this problem? Can u paste here some generated commented for a real-world function? so we can see the output? Maybe we should omit type information in these comments entirely and just use human-readable part of docs? 🤔

chmanie: This is actually not generating types, but only comments. Here's a real-life comment:
```
    /**
     * Downgrades are not allowed, i.e. `_newVersion` should be higher than the currect colony version.
     * Upgrades a colony to a new Colony contract version `_newVersion`.
     * @param _newVersion The target version for the upgrade
     */
    upgrade(
      _newVersion: BigNumberish,
      overrides?: TransactionOverrides
    ): Promise<ContractTransaction>;
```

Not sure if this is what you were asking.

krzkaczor: Oh perfect. I get it now. I thought it's gonna have type annotation as well. This is great! Can u proceed with implementation for rest of the targets? 

When done I'll merge it to `kk/new-architecture` branch and release new beta packages.

chmanie: Yup, I'll continue tomorrow!

chmanie:
One thing I noticed when trying out my branch with our contracts is that the command I was using

```
typechain --target ethers-v4 --outDir src/contracts lib/build/contracts/*.json
```

just matched the first file it could find, not all json files anymore. Has the command changed?

krzkaczor:
Some things changed in typechain core (like name normalization) but it should be fine. One thing: you miss quotes around glob pattern (otherwise shell interprets it). 

It the repo public, can I take a look if it still doesn't work?

chmanie:
> One thing: you miss quotes around glob pattern (otherwise shell interprets it).

🤦‍♀️That was it. It's all good now.

I pushed the adjustments for truffle-v4, truffle-v5 and web3-v1 (as we discussed I just copied and pasted the doc-generation function - it felt weird ;D)

Let me know if there's anything more to do

krzkaczor:
> it felt weird ;D

I knew it will 😆 

I will take a close look later today and merge. Thanks again! 

One other thing is that it would be cool to add natspec to contracts used for tests `contracts/` (at least to one method). So we can see it in the outputs. We use this as a sort of snapshot testing mechanics.

I am not sure if it's easy to force `solcjs` to generate combined artifacts (abi and natspec and more in one file). Currently it generated abi and bin totally separate.

krzkaczor:
Oh and btw. glob issue is releated to: https://github.com/ethereum-ts/TypeChain/issues/214

chmanie:
> One other thing is that it would be cool to add natspec to contracts used for tests contracts/ (at least to one method).

I'll take a look if I can integrate that.

Here's how to generate the documentation annotations in the ABI file as we interpret them: https://solidity.readthedocs.io/en/v0.5.12/natspec-format.html#documentation-output

chmanie:
Oh, I'm just seeing now that this project is using `solcjs` which doesn't support `--userdoc` or `--devdoc` options :(

krzkaczor:
Okay, what if we keep one special ABI + natspec file somewhere exactly for this. Can u paste here some example minimal file: like 1 function with, 1 natspec def but already compiled? and I ll take care of the rest.

chmanie:
Right, OK. This should do:

```
{
  "contractName": "SampleContract",
  "abi": [
    {
      "constant": false,
      "inputs": [
        {
          "name": "_wad",
          "type": "uint256"
        }
      ],
      "name": "mintTokens",
      "outputs": [],
      "payable": false,
      "stateMutability": "nonpayable",
      "type": "function"
    },
    {
      "constant": true,
      "inputs": [],
      "name": "version",
      "outputs": [
        {
          "name": "version",
          "type": "uint256"
        }
      ],
      "payable": false,
      "stateMutability": "pure",
      "type": "function"
    }
  ],
  "devdoc": {
    "title": "Contract interface",
    "methods": {
      "mintTokens(uint256)": {
        "params": {
          "_wad": "Amount to mint"
        }
      },
      "version()": {
        "return": "Version number"
      }
    }
  },
  "userdoc": {
      "mintTokens(uint256)": {
        "notice": "Mint `_wad` amount of tokens."
      },
      "version()": {
        "notice": "Get the Colony contract version. Starts from 1 and is incremented with every deployed contract change."
      }
    },
    "notice": "This is the contract's public interface"
  }
```

krzkaczor:
Published as:
```
typechain@2.0.0-beta2
```

and packages as:

```
x@1.0.0-beta1
```

Thanks again @chmanie

chmanie:
Fantastic! I shall try the beta branch in our project and report back!

PR #227, opened 2020-05-04
Use raw names for `artifacts.require`

JamesLefrere:
As far as I can tell, Truffle's `artifacts.require` should use the `rawName` rather than the normalized `name`. e.g. a contract named `ERC20` won't be found when `require`d as `Erc20`.


krzkaczor:
I think you're right, thanks! Before merging I want to add a test for this but I should be able to finish today.

krzkaczor:
Thanks James!

Published as:
```
@typechain/truffle-v4@2.0.0
@typechain/truffle-v5@2.0.0
```

(I identified this as a breaking change 🤔)

PR #228, opened 2020-05-04
Export contract types in artifact headers

JamesLefrere:
This PR changes the codegen for the Truffle artifact headers such that the contracts (and now, also contract instances) are exported and imported.

The use-case for this is the ease of importing multiple contracts from one location. I believe this was also the behaviour of TypeChain v1. Alternatively, this could also be done with a separate, optional file; but ideally this is done with TypeChain rather than in userland.

It seems as if the `no-unused-vars` eslint rule is violated in the generated files, so I'm not sure if there's a better solution here. `ts-generator` could also add `eslint-disable`, or maybe there's a neater solution.

### DIFF
--- a/packages/target-truffle-v4-test/types/truffle-contracts/index.d.ts
+++ b/packages/target-truffle-v4-test/types/truffle-contracts/index.d.ts
@@ -20,3 +20,13 @@ declare global {
     }
   }
 }
+
+export {
+  DataTypesInputContract,
+  DataTypesInputInstance
+} from "./DataTypesInput";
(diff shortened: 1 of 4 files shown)


JamesLefrere:
Oops, that's what I get for skipping tests.

krzkaczor:
Btw. I agree that `ts-generator` should generate `eslint-disable` comments as well.

krzkaczor:
Thanks James!

Published as:
```
@typechain/truffle-v4@2.0.0
@typechain/truffle-v5@2.0.0
```

PR #233, opened 2020-05-13
Fix incorrect return type for contract.at()

petejkim:
`at()` is an asynchronous function that returns a promise.

v5: https://github.com/trufflesuite/truffle/blob/693b413f3507ca4cf0fb9c1cb02ef04665b3a945/packages/contract/lib/contract/constructorMethods.js#L60

v4: https://github.com/trufflesuite/truffle/blob/v4.1.17/packages/truffle-contract/contract.js#L427


krzkaczor:
Thanks, makes sense!

krzkaczor:
Released as:
`@typechain/truffle-v5@2.0.1`
`@typechain/truffle-v4@2.0.1`

PR #237, opened 2020-05-20
docs: fix broken links under "Packages" section

PaulRBerg:
I assumed that they should be pointed to the local paths in the repo.


krzkaczor:
Upps, thanks for catching that!

PR #238, opened 2020-05-20
docs: fix broken commands and examples

PaulRBerg:
With "ethers" set as the target instead of "ethers-v4", I was getting this error:

```
Error occured:  Couldn't find ethers. Tried loading: typechain-target-ethers, @typechain/ethers, /Users/paulrberg/Projects/Sablier/Magellan/Protocol/packages/contracts/ethers.
Perhaps you forgot to install typechain-target-ethers?
```

This PR fixes the commands in the README to refer to "ethers-v4" instead of "ethers", and it also comes with a few formatting enhancements.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,8 +25,7 @@
 
 - static typing - you will never call not existing method again
 - IDE support - works with any IDE supporting Typescript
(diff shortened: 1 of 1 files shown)


krzkaczor:
Again, thank you for making TypeChain better! :)

PR #250, opened 2020-06-20
Add ethers v5 target

### LINKED ISSUE #205: Add target for ethers v5
I haven't gotten a good look at how v5 has changed in terms of typing, but I think additional work would be needed to support v5 as a target. It's in beta right now so no rush, but would like to get it on the roadmap :)

zemse:
This PR adds a `target-ethers-v5` to `TypeChain` packages and intends to close #205.

Thanks to @krzkaczor and @quezak for v4 target. The v5 target in this PR builds on top of it to update the types changed in ethers v5.

To try it: by cloning repository:

1. `git clone git@github.com:zemse/TypeChain.git` or `git clone https://github.com/zemse/TypeChain.git`
2. `cd TypeChain`
3. `yarn` (you need yarn installed if you don't)
4. `yarn build` (this generates dist dir which is imp)
5. Link this custom target to your project by placing the appropriate path to the target: `typechain --target ../TypeChain/packages/target-ethers-v5/` ([Docs](https://github.com/ethereum-ts/TypeChain#usage))

Just in case if it's too much, I have temporarily published it to npm for testing:

1. `npm install typechain-target-ethers-v5 --save-dev` (later you can just replace the package with official one)
2. `typechain --target ethers-v5` ([Docs](https://github.com/ethereum-ts/TypeChain#usage))

### Useage

```typescript
import { Erc20Factory } from '../../interfaces'; // there is an index.ts file which exports all factories

// Deploying
// you only need to pass a signer (abi and bytecode is managed)
const erc20Factory = new Erc20Factory(signer); 
const erc20Instance = await erc20Factory.deploy(); // you will get type suggestions in VSCode.

// Connecting to Existing
const erc20Instance = Erc20Factory.connect(contractAddress, signer);

// For some reason if you need interface of the contract
import { Erc20 } from '../../interfaces/Erc20';
interface Global {
  tokenInstance: Erc20;
}
```

Please have a look at this if it is fine and let me know of any change I need to make.


abarmat:
I tried this PR on a repo where I have many tests I migrated from ethers-v4 to ethers-v5 using waffle-v3 and worked for me.

.gitignore
```diff
@@ -14,3 +14,4 @@ examples/truffle-v4/types
 examples/truffle-v4/migrations
 examples/truffle-v5/types
 examples/truffle-v5/migrations
+.DS_Store
```
krzkaczor, line 17: this should go to your global gitignore file. otherwise we will end up with huge list of unrelevant OS files in this file.

krzkaczor:
This looks great @zemse! Thanks! 

I guess more tests would be nice (in a separate directory as we do for other targets).

packages/target-ethers-v5/src/codegen/index.ts
```diff
@@ -0,0 +1,258 @@
...
+  const _bytecode = "${bytecode.bytecode}";
+
+  ${generateLibraryAddressesInterface(contract, bytecode)}
+  `
+}
+
+export function codegenAbstractContractFactory(contract: Contract, abi: any): string {
+  return `
+  import { Contract, Signer } from "ethers";
+  import { Provider } from "ethers/providers";
```
abarmat, line 148: @krzkaczor is this ok, or it should be `import { Provider } from '@ethersproject/providers';` ?

HenryNguyen5:
It looks like this PR does not support https://github.com/ethers-io/ethers.js/issues/395#issuecomment-492701325, i think it would be really useful to add that in for those who use ethers to test their contracts too.

packages/target-ethers-v5/src/codegen/index.ts
krzkaczor: uhh good call. That's why we need tests + isolated example to catch such problems 😆

krzkaczor:
> It looks like this PR does not support staticCall, i think it would be really useful to add that in for those who use ethers to test their contracts too.

what the hell? it seems like v5 has two completely different calling conventions: https://docs.ethers.io/v5/migration/ethers-v4/#migration-v4--interface--functions

.gitignore
zemse: That makes sense. Removing this from gitignore file right away.

packages/target-ethers-v5/src/codegen/index.ts
zemse: Oh my bad.. fixing this for now. @krzkaczor I'll have a look at how other tests are written and add the same for v5 here.

zemse:
@HenryNguyen5 Just published `typechain-target-ethers-v5@1.2.1` with support for staticCall. Can you try if it works as expected? 

Maybe few more things might have been missed / incorrect. As we use them, we might find more. I'll go through the migration guide to find and add anything I might have missed.

zemse:
I'm not sure how `interface.encodeFunctionData("transfer", [ to, amount ])` can be added since the second argument's type is dependent on the first argument. However, an alternative [`populateTransaction`](https://github.com/ethers-io/ethers.js/issues/838#issuecomment-630514319) has been implemented.

Maybe we can keep this PR open for a week or two until we get enough feedback from more developers. Also tests are pending to be added, I'll be able to add them this weekend.

krzkaczor:
@zemse awesome work! Yeah I would like to hear some feedback from the users. Tests would also be great! We can't publish an official release without them.

> interface.encodeFunctionData("transfer", [ to, amount ])

I think it's possible to implement it by using function overloads where first arg type would be string literal `"transfer"`. But yes, it's a little bit more complicated.

zemse:
> I think it's possible to implement it by using function overloads where first arg type would be string literal "transfer". But yes, it's a little bit more complicated.

Okay, it's possible then I'll give it a try because it'd be nice to have this too.

quezak:
I'll take a look tomorrow!

packages/target-ethers-v5/src/codegen/types.ts
```diff
@@ -0,0 +1,85 @@
+import { EvmType, EvmOutputType, TupleType, AbiParameter, AbiOutputParameter } from 'typechain'
+
+export interface GenerateFunctionOptions {
+  returnResultObject?: boolean
+  isStaticCall?: boolean
+}
+
```
quezak, line 7: I'd move this interface to the `functions.ts` file, since it's only used there and doesn't have to be exported. (this file name `types.ts` refers to handling generation for solidity types, not TS types for this package ;) )

packages/target-ethers-v5/src/codegen/index.ts
```diff
@@ -0,0 +1,268 @@
...
+
+export function codegenContractTypings(contract: Contract) {
+  const contractImports: string[] = ['Contract', 'ContractTransaction']
+  const allFunctions = values(contract.functions)
+    .map(codegenFunctions.bind(null, { returnResultObject: true }))
+    .join('')
+
+  if (allFunctions.match(/\W Overrides(\W|$)/)) contractImports.push('Overrides')
+  if (allFunctions.match(/\WPayableOverrides(\W|$)/)) contractImports.push('PayableOverrides')
+  if (allFunctions.match(/\WCallOverrides(\W|$)/)) contractImports.push('CallOverrides')
```
quezak, line 21: why is the first regex different than the 2nd & 3rd one? (space between `\W` and the name)

packages/target-ethers-v5/src/codegen/index.ts
```diff
@@ -0,0 +1,268 @@
...
+
+    interface: ${contract.name}Interface;
+
+    functions: {
+      ${values(contract.functions)
+        .map(codegenFunctions.bind(null, { returnResultObject: true }))
+        .join('\n')}
+    };
+
+    ${values(contract.functions).map(codegenFunctions.bind(null, {})).join('\n')}
```
quezak, line 65: Does this (different options to `codegenFunctions` here) mean that in ethers v5 `contract.someMethod(...)` and `contract.functions.someMethod(...)` return a different type?

packages/target-ethers-v5/src/codegen/index.ts
```diff
@@ -0,0 +1,268 @@
...
+
+export function codegenContractFactory(contract: Contract, abi: any, bytecode?: BytecodeWithLinkReferences): string {
+  const constructorArgs =
+    (contract.constructor && contract.constructor[0] ? generateInputTypes(contract.constructor[0].inputs) : '') +
+    `overrides?: ${contract.constructor[0]?.stateMutability === 'payable' ? 'PayableOverrides' : 'Overrides'}`
+  const constructorArgNamesWithoutOverrides =
+    contract.constructor && contract.constructor[0] ? generateParamNames(contract.constructor[0].inputs) : ''
+  const constructorArgNames = constructorArgNamesWithoutOverrides
+    ? `${constructorArgNamesWithoutOverrides}, overrides || {}`
+    : 'overrides || {}'
```
quezak, line 106: does the `ContractFactory` require passing empty overrides, since you added the explicit `|| {}` here?

packages/target-ethers-v5/src/codegen/index.ts
quezak: maybe is it worth extracting a helper fn to help readability and avoid typos in the regexes? like
```typescript
function pushImportIfUsed(importName: string, generatedCode: string, importArray: string[]): void {
  if (new RegExp(`\\W${importName}(\\W|$)`).test(generatedCode)) importArray.push(importName);
}
```

zemse: I was afraid that `/\WOverrides(\W|$)/` would match `PayableOverrides` and `CallOverrides` too when it shouldn't and should only match `Overrides`. Is there a better way to do the same? I think would face the same issue having this as a method as the `pushImportIfUsed` example

packages/target-ethers-v5/src/codegen/types.ts
zemse: Very much thanks for the review. This makes sense and I'll make this change.

packages/target-ethers-v5/src/codegen/index.ts
zemse: Yes! I think that's not documented yet. But you can find it here https://github.com/ethers-io/ethers.js/issues/869

packages/target-ethers-v5/src/codegen/index.ts
zemse: When we don't pass any override using typechain's overrides, the `overrides` is set as `undefined`. But it looks like ethers v4 was fine with undefined `overrides` value but in v5 it throws an error. We can make it not pass an undefined override (+1 or 2 lines) or just fall-backing to an empty override as done corrently. Is it good doing like this?

packages/target-ethers-v5/src/codegen/index.ts
quezak: I've noticed that problem when I wrote those regexes in the `ethers-v4` target :) `\W` means a non-word character, and it's there specifically so `xyz` doesn't match `abcxyz`, so it's safe to drop the extra space.

quezak: Example: https://regex101.com/r/r6Cjn4/1

packages/target-ethers-v5/src/codegen/index.ts
quezak: This is weird, not passing and optional argument _should_ be handled exactly the same as explicitly passing `undefined` there. TBH, if passing `undefined` overrides throws in ethers v5, it's likely a bug there. We can leave the explicit empty overrides for now.

zemse:
> I think it's possible to implement it by using function overloads

Thanks for pointers. I've tried to add that.

I've updated the tests for ethers v5. Please have a look and let me know for any changes required.

krzkaczor:
Thanks for all your hard work! It's moved to internal branch `fork/ethers-v5-target`. For now.

krzkaczor:
It's published as `@typechain/ethers-v5`, still in beta. Thanks again for your work. 

I added an example to my branch and it seems like everything works fine 👍

zemse:
@krzkaczor Thanks for moving it forward! I'm very happy I gave this a shot, this was my first PR in OSS. I have marked the demo published package as deprecated. Let me know if I need to do anything more.

BTW saw the typedAssert while writing tests, It might be common for rest of you guys but for someone like me without a programing background and barely over 1 year exp, honestly I'm was impressed with the cool stuff that I saw in this repo.

One more thing which were exiting for me were the reviews (thanks to everyone for taking time to review the work and guiding me). Sorry if I sound too crazy, but I'm happy that I'm on GitHub, I have access to so many amazing people all around the world. I observed that in preparing PRs on GitHub, we get to learn a lot of best practices since we get to see a lot. So I believe I'll find more opportunities to do.

Hopefully, I'll get to see you all again.

abarmat:
@zemse thank you for your time putting this code together. This PR is super useful for all us using Typechain and Ethers.

zemse:
Hi @HenryNguyen5, I just encountered a problem with `staticCall`. https://github.com/ethers-io/ethers.js/issues/924

krzkaczor:
@zemse wow great work! Especially that it was your first PR to OSS.

I gotta admit that TypeChain is not the easiest project to contribute to, so you should be proud of yourself even more ;)

PR #252, opened 2020-06-30
Fork/ethers v5 target

krzkaczor:
Published as `@typechain/ethers-v5`


zemse:
If https://github.com/ethers-io/ethers.js/issues/924 resolves into finalizing `callStatic`, this target will need to be updated. But if `ethers.js` changes `callStatic` to `staticCall`, then nothing is required to do in this case.

bradennapier:
When trying this target, I noticed that the `estimateGas` function does not allow transaction details.  It looks like not too long ago you just allowed `estimateGas` to be inherited from `Contract` which defines functions as `(...args: any[]) => Promise<BigNumber` IIRC. 

When target is truffle-v5:

```typescript
depositEther: {
      (txDetails?: Truffle.TransactionDetails): Promise<
        Truffle.TransactionResponse<AllEvents>
      >;
      call(txDetails?: Truffle.TransactionDetails): Promise<void>;
      sendTransaction(txDetails?: Truffle.TransactionDetails): Promise<string>;
      estimateGas(txDetails?: Truffle.TransactionDetails): Promise<number>;
    };
```

However, the `estimateGas` when ethers:

```typescript
depositEther(): Promise<BigNumber>;
```

This seems to be potentially related to https://github.com/ethereum-ts/TypeChain/pull/252/commits/c759bad6daa47b50a22ebfad1ae7f251ac973e19 since this fn in the contract is `payable`. 

It does generate the overrides for the `payable` outside of the `estimateGas`

```typescript
depositEther(overrides?: PayableOverrides): Promise<ContractTransaction>;
```

bradennapier:
Created PR to fix this problem: https://github.com/ethereum-ts/TypeChain/pull/254

Also about to open the PR to fix ethers-v4

packages/target-ethers-v5/src/index.ts
```diff
@@ -0,0 +1,132 @@
...
+    }
+  }
+
+  afterRun(): TFileDesc[] {
+    // For each contract that doesn't have bytecode (it's either abstract, or only ABI was provided)
+    // generate a simplified factory, that allows to interact with deployed contract instances.
+    const abstractFactoryFiles = Object.keys(this.contractCache).map((contractName) => {
+      const { contract, abi } = this.contractCache[contractName]
+      return {
+        path: join(this.outDirAbs, `${contract.name}Contract.ts`),
```
quezak, line 119: @krzkaczor this looks invalid, the file should still be named `...Factory.ts`, since it still exports `class ${contract.name}Factory`.
@zemse was that a mistake or did you change this intentionally?
```suggestion
        path: join(this.outDirAbs, `${contract.name}Factory.ts`),
```

(after the change we'll need to re-generate the files in `packages/target-ethers-v5-test/types/` since they include the invalid filename)

zemse: You're correct, there's a mistake here, this should be `${contract.name}Factory.ts`. I'll create a PR to fix this.

quezak:
BTW: I'm running the released `"@typechain/ethers-v5": "^0.0.3",` locally for a few days now and found no problems after patching the factory filenames.

krzkaczor:
I just published: `@typechain/ethers-v5@0.0.4`

@quezak yes, I think it's reasonable to publish it as stable this weekend 👍

krzkaczor:
Finally, published as: `@typechain/ethers-v5@1.0.0`.

Thanks for everyone involved in this! Especially @zemse and @bradennapier!

PR #254, opened 2020-07-09
Fix target-ethers-v5 to allow transaction and contract call overrides 

bradennapier:
> **Note:** A lot of these changes appear to be due to the fact that I called `yarn test:fix` which the documentation for contributing specifies to do before pushing.  
> 
> I only changed two files in this: [codegen/index.ts](https://github.com/ethereum-ts/TypeChain/compare/fork/ethers-v5-target...bradennapier:fix/ethersv5-fn-types?expand=1#diff-eba1dc703ae02ade2901088a1a8fd63a) and [codegen/functions.ts](https://github.com/ethereum-ts/TypeChain/compare/fork/ethers-v5-target...bradennapier:fix/ethersv5-fn-types?expand=1#diff-41133d8d9ebf5a20059fc427efcc7fb0)

According to https://docs.ethers.io/v5/api/contract/contract/#Contract--check the signature of the `estimateGas` , `populateTransaction` and others should allow for overrides.  These are not currently being built making it impossible to call them in many cases. 

This simply reproduces the logic done for the functions which are produced correctly by adding the overrides (and documentation) for those calls.

> It may be ideal to modify the documentation to mention it is an `estimateGas` etc but I didn't want to make that change myself.

> I noticed shortly after this that the `ethers-v4` is also affected by this flaw.  We used the `ethers-v4` typechain previously to produce for `ethers-v5` and just made some modifications to make it work.  When we did this it was defaulting `estimateGas` to the ethers definition which allows any arguments.

> When using overloads there is a slight bug that should be addressed (probably in another PR).  When you use overloads, [the first function defined should be defined as the name of the fn](https://github.com/ethers-io/ethers.js/issues/407#issuecomment-458329708), then all overloads should have the function names.  

Given the ERC-20 example which is better types by ethers: https://docs.ethers.io/v5/api/contract/example/#example-erc-20-contract--meta-class-methods

#### Before

```typescript
estimateGas: {
    name(): Promise<BigNumber>;
    symbol(): Promise<BigNumber>;
    decimals(): Promise<BigNumber>;
    totalSupply(): Promise<BigNumber>;
    balanceOf(account: string): Promise<BigNumber>;
    transfer(recipient: string, amount: BigNumberish): Promise<BigNumber>;
    allowance(owner: string, spender: string): Promise<BigNumber>;
    approve(spender: string, amount: BigNumberish): Promise<BigNumber>;
    transferFrom(
      sender: string,
      recipient: string,
      amount: BigNumberish
[...]

PR #255, opened 2020-07-09
Fix target-ethers-v4 to allow transaction and contract call overrides 

bradennapier:
Similar to https://github.com/ethereum-ts/TypeChain/pull/254 - this fixes the situation to provide override arguments as specified.  

There are a few specifics that are considered here:

- According to [the documentation](https://docs.ethers.io/v4/api-contract.html#overrides) for ethers v4 (and my own experience where i needed these overrides), **every** method has overrides available to it.  Their example only shows a few properties but when looking at the v5 types they are defined extending the transaciton overrides while adding a `from` key (and `blockTag` which im not sure, i didnt add that). 

- I did not add the differentiation that the `target-ethers-v5` has where it checks if the fn is a constant function - this is due to the fact that it seemed to be doing some extra changes to handle that properly and I didn't want this to get too big of a change.

- I added some tests to call both the contract and the `contract.functions` method - although not testing the overrides being used as I wasn't sure the best way to implement that with y;alls setup.  I just needed this to be working right for our project  :-)

- When using overloads there is a slight bug that should be addressed (probably in another PR).  When you use overloads, [the first function defined should be defined as the name of the fn](https://github.com/ethers-io/ethers.js/issues/407#issuecomment-458329708), then all overloads should have the function names.  

---

The 3 PR's combined, which we are using to generate working types for our contracts is https://github.com/bradennapier/TypeChain/tree/combined-prs

### DIFF
--- a/packages/target-ethers-v4-test/test/Overload.test.ts
+++ b/packages/target-ethers-v4-test/test/Overload.test.ts
@@ -16,11 +16,16 @@ describe('Overloads', () => {
   afterEach(() => ganache.close())
 
   it('works with 1st overload', async () => {
-    const result = await contract.functions['overload1(int256)'](1)
+    const result = await contract['overload1(int256)'](1)
     typedAssert(result, new BigNumber(1))
   })
 
   it('works with 2n overload', async () => {
+    const result = await contract['overload1(uint256,uint256)'](1, 2)
+    typedAssert(result, new BigNumber(3))
+  })
+
+  it('works when calling the contract.functions version', async () => {
     const result = await contract.functions['overload1(uint256,uint256)'](1, 2)
     typedAssert(result, new BigNumber(3))
   })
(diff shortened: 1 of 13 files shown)


krzkaczor:
@quezak thoughts on this? It seems like it's a breaking change for some cases, gosh these overloads are such a PITA.

quezak:
I'll try to check it over the weekend

bradennapier:
I'm not sure how it would be a breaking change? It should only be adding an extra optional property. 

We are using this fork way more than the v5 changes so it's much more heavily tested.  Our backend is all v4 and only our front end is v5.

krzkaczor:
Sorry probably GH diff combine changes from both of your PRs. Yeah probably the other PR introduced that change, but anyway.

I was referring to this change: https://github.com/ethereum-ts/TypeChain/pull/255/files#diff-72918e744bb576e96e0347956777735cR63

Before, calling `overload1` was allowed.

bradennapier:
Hmm yeah on v4 it should allow that - maybe when i fixed that didnt get to this branch.  v5 is the one that changed that behavior significantly and removes the ability all together

bradennapier:
Ah yes - https://github.com/ethereum-ts/TypeChain/pull/256 this fixes it

krzkaczor:
Published (with additional fixes) as: `@typechain/ethers-v4@1.0.1`

PR #256, opened 2020-07-09
fix: overloads to work according to spec

bradennapier:
According to the discussions in [this topic](https://github.com/ethers-io/ethers.js/issues/407) overloads work a bit differently than they were being typed:

- in `ethers-v4` overloads will use the first defined as the normal fn (`contract.overload1(1)`) and will ALWAYS have `contract['overload1(int256)'](1)`
- in `ethers-v5` overloads will **only** provide the latter if they are detected so that `contract.overload1(1)` is not available.  

I added some extra testing to confirm this was the appropriate behavior as well.  

> i added a case for `ethers-v5` that will continually confirm they have not changed back to the `v4` behavior by expecting the `contract.overload1` to be undefined explicitly. 

> Additionally, I created tests for both v4 and v5 which tests against and confirms the behavior differences -- in v4 `contract.functions.method()` returns the same as a normal `contract.method()` but v5 returns `[Result]` for `contract.functions.method()` 

This should start working properly when combined with my other PR's as well for the other areas, but work adding some tests once they are all a family :-)

---

The 3 PR's combined, which we are using to generate working types for our contracts is https://github.com/bradennapier/TypeChain/tree/combined-prs

### DIFF
--- a/packages/target-ethers-v4-test/test/DataTypesInput.test.ts
+++ b/packages/target-ethers-v4-test/test/DataTypesInput.test.ts
@@ -35,6 +35,11 @@ describe('DataTypesInput', () => {
       '0x70b144972C5Ef6CB941A5379240B74239c418CD4',
     )
 
+    typedAssert(
+      await contract['input_address(address)']('0x70b144972C5Ef6CB941A5379240B74239c418CD4'),
+      '0x70b144972C5Ef6CB941A5379240B74239c418CD4',
+    )
+
     typedAssert(await contract.input_bytes1('0xaa'), '0xaa')
     typedAssert(await contract.input_bytes1([0]), '0x00')
 
--- a/packages/target-ethers-v4-test/test/Overload.test.ts
+++ b/packages/target-ethers-v4-test/test/Overload.test.ts
@@ -16,12 +16,20 @@ describe('Overloads', () => {
   afterEach(() => ganache.close())
 
   it('works with 1st overload', async () => {
-    const result = await contract.functions['overload1(int256)'](1)
-    typedAssert(result, new BigNumber(1))
+    const results = await Promise.all([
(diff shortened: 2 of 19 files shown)


krzkaczor:
Thanks for this @bradennapier I want to finally merge this, the plan is:
1. I already merged https://github.com/ethereum-ts/TypeChain/pull/255
2. Resolve conflicts and merge this. 
3. Merge whole `fork/ethers-v5-target` to master.
4. Republish packages.

bradennapier:
Sounds good, yep.  Let me know if any questions - I think this one is actually necessary for #255 to work properly but I do think there were some conflicts that had to be fixed so lmk if any confusion there.  My linked branch that fixed them should help as well.  

Happy to see it all coming in.  Still don't use overloads, lol - but the tests should confirm its all legit. <3

krzkaczor:
@bradennapier thanks! Actually I run into a problem while fixing conflicts. 

Could you take a look? it seems like tests are failing after regerating types. Seems like default override is missing? `contract.functions.overload1(1),`. I just did a merge commit with `fork/ethers-v5-target` and run `yarn test` to regenerate types (they were the source of conflicts) and then `yarn test:fix` for final test before merge.

Also, in the meantime I improved a little bit testing pipeline so maybe that's why it wasn't caught before.

bradennapier:
v5 works oddly.  If there are any overloads it actually doesn't have the default overload, only the string version.

krzkaczor:
It fails actually in v4 tests. Note I merged your previous PR (#255) to master, not to this branch. Is this correct?

bradennapier:
Closing in favor of https://github.com/ethereum-ts/TypeChain/pull/269 which incorporates conflict fixes and confirmed working with tests

PR #258, opened 2020-07-12
Fix callStatic overrides for non constant fns

zemse:
Before:

```typescript
callStatic: {
    getValue(overrides?: CallOverrides): Promise<string>;
    setValue(overrides?: Overrides): Promise<string>;
}
```

After:

```typescript
callStatic: {
    getValue(overrides?: CallOverrides): Promise<string>;
    setValue(overrides?: CallOverrides): Promise<string>;
}
```

In `CallOverrides` we can pass `from` property, which can be passed to non constant methods under `callStatic` bucket. Until now only constant methods were having the `CallOverrides` overrides. This PR fixes the specific case bug.

Testing: I've ran this PR in two of my projects with no breakings. However, to be more sure, can someone try in their project? It'd be great.

### DIFF
--- a/packages/target-ethers-v5/src/codegen/functions.ts
+++ b/packages/target-ethers-v5/src/codegen/functions.ts
@@ -26,7 +26,7 @@ function generateFunction(options: GenerateFunctionOptions, fn: FunctionDeclarat
   return `
   ${generateFunctionDocumentation(fn.documentation)}
   ${overloadedName ?? fn.name}(${generateInputTypes(fn.inputs)}${
-    !isConstant(fn) && !isConstantFn(fn)
+    !options.isStaticCall && !isConstant(fn) && !isConstantFn(fn)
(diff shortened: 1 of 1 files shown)


zemse:
@krzkaczor This PR introduces a bug when `CallOverrides` is not imported (in a case when there are only non-constant fns in a contract).

While there is a contract `Payable.sol` in the test cases that could expose the problem, I am surprised why typescript is not giving error in TypeChain project. To reproduce `yarn test` and go to https://github.com/ethereum-ts/TypeChain/blob/fork/ethers-v5-target/packages/target-ethers-v5-test/types/Payable.d.ts#L72-L76. The `CallOverrides` are interpreted as `any` type. This caused the CI to pass, when it should have failed. Can you please have a look why it doesn't error?

In the meantime, I'll fix and create a PR.

krzkaczor:
@zemse thanks for this comment, turns out that "no git changes" rule wasn't enforced. Recently changed files weren't committed and thus weren't typechecked. Now CI fails on this branch as expected :)

PR #265, opened 2020-07-24
Fix abstract contract filename

zemse:
This PR fixes the review https://github.com/ethereum-ts/TypeChain/pull/252#pullrequestreview-454800523


krzkaczor:
Thanks!

PR #274, opened 2020-08-15
Support fixed length arrays

### LINKED ISSUE #273: Better typing for fixed length array arguments (ethers v5 target)
```solidity
function payToTree(uint32[3] _rewardRatio) public;
```
Current:
```ts
_rewardRatio: ethers.BigNumberish[]
```

Desired:
```ts
_rewardRatio: [ethers.BigNumberish, ethers.BigNumberish, ethers.BigNumberish]
```

I think it would be nice to take into account about the array length (since ethers.js's ABI encoder also takes it into account https://github.com/ethers-io/ethers.js/issues/1004).

I can work on this later this month or may be next month. Though, in case anyone is interested may do it :)

zemse:
Closes #273

### DIFF
--- a/packages/target-ethers-v5-test/types/DataTypesInput.d.ts
+++ b/packages/target-ethers-v5-test/types/DataTypesInput.d.ts
@@ -62,7 +62,7 @@ interface DataTypesInputInterface extends ethers.utils.Interface {
   ): string;
   encodeFunctionData(
     functionFragment: "input_stat_array",
-    values: [BigNumberish[]]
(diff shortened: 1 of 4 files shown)


krzkaczor:
This is a great addition, thanks!

I want to release it as breaking change, I think it might break some of the existing code... WDYT?

zemse:
I have written few test cases using my local typechain branch target. I just regenerated types using the released target and didn't get any ts errors in the tests. 

By this, I assume that this might not be a breaking change but instead it could be a bug fix, since the code would anyways throw on runtime. Can this be a correct assumption?

Issue #278, opened 2020-09-10
[ethers-v5] Contract types not exported in generated index.ts


tuler:
The index.ts only reexports the factory types, it does not export the contract types, the ones at `<ContractName>.d.ts`.
These types should also be exported.

zemse:
I did try this in my PR but faced some typescript errors.

I had two ways of doing this:

#### 1. Export both definations and factories from `index.ts`

Issue: Typescript doesn't complain at all. `ts-node` throws an error (in my project which consumes typechain and runs mocha tests with ts-node) while requiring these files:
```
/project-path/node_modules/ts-node/src/index.ts:586
          throw new TypeError(
                ^
TypeError: Unable to require file: build/typechain/ERC20.d.ts
This is usually the result of a faulty configuration or import. Make sure there is a `.js`, `.json` or other executable extension with loader attached before `ts-node` available.
```

#### 2. An attempt to fix faulty export I tried exporting definitions from `index.d.ts` and factories from `index.ts`

Issue: Now typescript complains because it tries to look in `index.ts` and ignores `index.d.ts` for the contract types.

So I gave up for this one while submitting PR #250 thinking to look at later. @krzkaczor, anything we can do to do this or could this be a bug in `ts-node`? Since typescript doesn't error for the first case, I think this can be done someway.

tuler:
@zemse do you still have a branch with that try hanging around?
How did you try the re-export of option 1?

It should be:
```
export * from "./<ContractName>";
```

not
```
export * from "./<ContractName>.d.ts";
```

zemse:
I was using `export * from "./<ContractName>.d";`. Pushed it [here](https://github.com/zemse/TypeChain/blob/278-add-type-exports/packages/target-ethers-v5/src/index.ts#L129).

I tried with `export * from "./<ContractName>";`, it errors `Error: Cannot find module './<ContractName>'`

zemse:
> `export * from "./<ContractName>";`

Does this line in your `index.ts` file work in your project repo?

tuler:
This change works for me.
```diff
-  .map((fileName) => [`export * from './${fileName}Factory'`, `export * from './${fileName}.d'`].join('\n'))
+  .map((fileName) => [`export * from './${fileName}Factory'`, `export * from './${fileName}'`].join('\n'))
```

But there is a catch. I know it only makes sense to export factories of contracts which are concrete, excluding abstracts and interfaces.

[...]

PR #279, opened 2020-09-27
[ethers-v5] Export contract types in index.ts

zemse:
#278

Before:
```
<Contract>.d.ts
<Contract>Factory.ts
```

Only Factory file was exported

After:
```
<Contract>.ts
<Contract>Factory.ts
```

Now both the ts files are exported from `index.ts`

I've tried this PR on my current project, and didn't get any breakings. Though it can be tried on multiple projects to be sure.

### DIFF
--- a/packages/target-ethers-v5/src/codegen/index.ts
+++ b/packages/target-ethers-v5/src/codegen/index.ts
@@ -132,7 +132,7 @@ export function codegenContractFactory(contract: Contract, abi: any, bytecode?: 
   import { Provider, TransactionRequest } from '@ethersproject/providers';
   import { ${ethersContractImports.join(', ')} } from "@ethersproject/contracts";
 
-  import { ${contract.name} } from "./${contract.name}";
+  import type { ${contract.name} } from "./${contract.name}";
 
   export class ${contract.name}Factory extends ContractFactory {
     ${generateFactoryConstructor(contract, bytecode)}
@@ -166,7 +166,7 @@ export function codegenAbstractContractFactory(contract: Contract, abi: any): st
   import { Contract, Signer } from "ethers";
   import { Provider } from "@ethersproject/providers";
 
-  import { ${contract.name} } from "./${contract.name}";
+  import type { ${contract.name} } from "./${contract.name}";
 
   export class ${contract.name}Factory {
     static connect(address: string, signerOrProvider: Signer | Provider): ${contract.name} {
--- a/packages/target-ethers-v5/src/index.ts
+++ b/packages/target-ethers-v5/src/index.ts
@@ -125,7 +125,14 @@ export default class Ethers extends TsGeneratorPlugin {
       ...abstractFactoryFiles,
       {
         path: join(this.outDirAbs, 'index.ts'),
-        contents: this.contractFiles.map((fileName) => `export * from './${fileName}Factory'`).join('\n'),
+        contents: this.contractFiles
+          .map((fileName) =>
+            [
+              `export { ${fileName}Factory } from './${fileName}Factory'`,
+              `export type { ${fileName} } from './${fileName}'`,
+            ].join('\n'),
+          )
+          .join('\n'),
       },
     ]
   }


zemse:
I was thinking why even we are having 2 files, when we can have just one?

```
<Contract>.ts // exports both Contract abstract class and ContractFactory class
```

quezak:
1. One big question first: why did you mark the generated class as abstract? It's not abstract, since it's just a more strictly typed extension of the concrete `Contract` class exported from Ethers.

2. Re `.ts` files for contract classes: I'm for it, but maybe @krzkaczor will oppose since he initially created this generator with just the type declarations being generated :) One reason was to avoid adding runtime overhead, but IMO that's not a big issue. And having only type declarations actually prevented me a few times to do useful tricks with the generated classes, for example passing a contract subclass to a helper function, or using `contractInstance.constructor.name`. One potential problem here are the factories: because `connect` just returns `super.connect`, the resulting object will be a plain `Contract` instance -- but if we'd add concrete `.ts` files for contract classes, it should produce instances of that class, so the factories should probably be updated.

3. Re 2 separate files -- I think that's OK, it's a common convention that `Xyz` class is exported from `xyz.ts`. Another reason to not merge the 2 files is that the factory file can be long, since it contains the contract ABI JSON.

zemse:
### Marking abstract

I thought it this way: If we see `Contract` from typescript's eyes (assuming a erc20 contract instance), then according to typescript there is no such `transfer` method (assuming the [wild card](https://github.com/ethers-io/ethers.js/blob/d817416bae2fbc7adb8391fd038613241b7ab8ba/packages/contracts/src.ts/index.ts#L572) is not there). But we know that infact a method exists there in javascript level, but whose implementation is not in the typescript. So typescript would allow two ways working with this abstract class:

1. Inherit it in other class and implement the methods in typescript level (which we dont do)
2. Have a function that returns an object that implements these methods in JS level and put return type of this abstract class. (this is what the contract factory is).

The second point might be my misconception, since type definitions (`.d.ts`) files are for this purpose. But I thought it could be fine to use abstract classes as a replacement for type definations.

### Re .ts files for contract classes

I tried a lot of ways to make everything export from `index.ts` in the past time when I was making the first PR. But all my attempts failed. I finally gave up with only exporting the factories. Now, this is the only way I have found yet that actually doesn't error when run in my project. Importing `d.ts` files in other files result in an error `module not found`.

> I'll try to run the updated code anyway tomorrow and see the results.

That'd be really great!

### Re 2 separate files 

Ok actually that's not a problem

krzkaczor:
First of all thanks @zemse for working on this!

Can't we just keep changes in this PR minimal to fix #278 ? I always preferred for TypeChain to generate pure types as this simplifies things and has no runtime impact. Then we don't need to debate on use of `abstract` keyword either ;) 

To make it work you might need `export type`: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html#type-only-imports-and-export

zemse:
Okay, thanks for giving the pointer for Type-Only Imports and Exports! By looking at the docs, I think this should solve the problem.

zemse:
Hi @krzkaczor I think there is some problem in `ts-generater` when using type imports:

```ts
Error occured:  '=' expected. (8:15)
   6 |   import { Contract, ContractFactory, Overrides } from "@ethersproject/contracts";
   7 | 
>  8 |   import type { ContractName } from "./ContractName";
     |               ^
   9 | 
  10 |   export class ContractNameFactory extends ContractFactory {
  11 |     
npm ERR! code ELIFECYCLE
npm ERR! errno 1
```

This problem comes when generating typechain files in the project.

krzkaczor:
@zemse yeah it seems like `ts-generator` needs to be updated and `prettier` version bumped. It will be a breaking change regarding formatting. Can you just commit all your work for this PR and I will take care of fixing this problem?

zemse:
The changes are minimal, but I fear it might not work due to further some error, and might need to be tweaked

krzkaczor:
Merged to tmp branch for now

Issue #216, opened 2020-02-17
Fails to generate types with a library that defines enums


stwiname:
When I attempt to generate types I am getting this error `Error occured:  Unknown type: ALib.BOOL`

I have attached a minimal example: [Archive.zip](https://github.com/ethereum-ts/TypeChain/files/4212472/Archive.zip)
To repro: `yarn && yarn build`

The error seems to be the same as https://github.com/ethereum-ts/TypeChain/issues/125 but is still happening

quezak:
Preliminary research shows that the ABI for the `other` function is different for a library:
```jsonc
// ...
      "inputs": [
        {
          "internalType": "enum ALib.BOOL",
          "name": "b",
          "type": "ALib.BOOL"
        }
```

While if I just change `interface ALib` to `contract ALib`, the type of the argument is plain `uint8`:
```jsonc
// ...
      "inputs": [
        {
          "internalType": "enum ALib.BOOL",
          "name": "b",
          "type": "uint8"
        }
```

I'll check if that's only when compiling with Truffle (personally I'm using Waffle :) )

quezak:
Confirmed, the situation is the same. We'll need to see why does the ABI differ between these two cases and how can we handle that in the generation step -- it will probably affect all targets.

Btw, sorry @stwiname for the delayed response.

SilentCicero:
I'm stuck on this as well.

quezak:
@SilentCicero what solc version are you using? I'm wondering if it's still the same in v0.7.0

SilentCicero:
6.8 currently with buidler

krzkaczor:
@stwiname @SilentCicero Can someone prepare minimal repro for this?

quezak:
@krzkaczor OP already contains a minimal repro :D when I checked it in March it was correct (ie. was bugged as described [above](https://github.com/ethereum-ts/TypeChain/issues/216#issuecomment-606583020))

krzkaczor:
Sorry I missed it 😆 I will work on this later today

krzkaczor:
This should be fixed now! Relased as `typechain@2.0.1`

SilentCicero:
Sweet, thanks all!

cameel:
[...]

PR #281, opened 2020-10-01
Support enums in libs

krzkaczor:
Enums in libs are such a mess... Basically, info from ABI is almost useless because it lacks exact information about enum type definition. The workaround is to detect enum type by using `internalType` hint and then replacing it with `uint8`. This is not supported by `truffle v4` compiler but I don't really care 😆 

Also I needed to hand tweak `yarn.lock` because of some mocha overrides problems.

Closes https://github.com/ethereum-ts/TypeChain/issues/216


quezak (approved):
Code looks OK :ok_hand: (but I didn't run it myself, maybe the guys who reported the issue will be willing to test?)

Should we maybe add some info to truffle target readme that it doesn't support library enums?

packages/typechain/src/parser/parseEvmType.ts
```diff
@@ -89,5 +89,9 @@ export function parseEvmType(rawType: string, components?: EvmSymbol[]): EvmType
     return { type: 'bytes', size: parseInt(match![1] || '1'), originalType: rawType }
   }
 
+  if (internalType?.startsWith('enum')) {
+    return parseEvmType('uint8')
+  }
```
cameel, line 94: Enums aren't always encoded as `uint8`. From [Solidity ABI spec](https://solidity.readthedocs.io/en/latest/abi-spec.html#mapping-solidity-to-abi-types):
> | Solidity | ABI |
> |---|---|
> | `enum` | smallest uint type that is large enough to hold all values<br><br>For example, an enum of 255 values or less is mapped to uint8 and an enum of 256 values is mapped to uint16.

Changing this is currently being discussed (see https://github.com/ethereum/solidity/issues/10035 and [Solidity Language Design: Enums with max 256 values](https://groups.google.com/forum/#!msg/solidity-users/Y2DrM_dyBf4/e_jwFMATAwAJ)), but just FYI, very large enums that map `uint16` are currently possible.

krzkaczor: Thanks for this note, I was afraid that this assumption may break. I think it's gonna be just fine in our case as it's rather there just to make parsing happy...

quezak: we can add a note in a comment there, just so no one is surprised or spends half a day resaarching if someone actually tries to use such a large enum :D

krzkaczor: Yup, I was planning on fixing it while working on TC today.

PR #286, opened 2020-10-08
[Ethers-v5] Export types

krzkaczor:
Closes: https://github.com/ethereum-ts/TypeChain/issues/278


krzkaczor:
@zemse can you test in on your codebase before I release? I will mark it as major update since formatting WILL change (breaking changes in prettier). I think it's the only sane thing to do...

zemse:
I used this on the code base, it works fine including the exporting types from index.ts. 

I think it can be worth mentioning that projects that use this new version need `typescript>=3.8`. New projects would have that but old projects will need to upgrade their typescript.

krzkaczor:
@zemse thanks for testing! I added some more fixes and gonna release it in a sec.

PR #289, opened 2020-10-27
Bump ethers to 5.0.19

tuler:
Updates ethers dependency to latest version at the moment.

### DIFF
--- a/examples/ethers-v5/package.json
+++ b/examples/ethers-v5/package.json
@@ -18,6 +18,6 @@
     "typescript": "^3.8.3"
   },
   "dependencies": {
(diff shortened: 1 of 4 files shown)


quezak:
Normally we should allow using a lower but compatible version, but IIRC there were some internal packaging fixes in ethers along these versions, which may cause the import problems some people reported.

I think it's OK, I'll have to test the update on my project first.

krzkaczor:
Thanks @tuler! This is fixed as part of: https://github.com/ethereum-ts/TypeChain/pull/289 It's not really bumped in target but rather non-peer dependency is removed which should be enough to fix it.

Issue #234, opened 2020-05-15
Typechain changing file names


PhABC:
I just upgraded from `typechain@1.03` to `typechain@2.0.0` and the generated file had a different name, which is unfortunate as I would have to modify many files to adapt to that change. Here are some examples of files that were affected:

`contracts/IERC20.d.ts` is now `contracts/Ierc20.d.ts`
`contracts/ERC1155.d.ts` is now `contracts/Erc1155.d.ts`

Any config I can setup such that the name of these type files is the same as the abi's name? Would be simpler.

petejkim:
I've run into something similar.

I had a name like `FooV1_1` where `V1_1` signifies `v1.0`, but TypeChain changes the name to `FooV11` (v11) :(

Could you make it preserve original naming please?

krzkaczor:
Hey @PhABC 

Yeah, it's one of the breaking changes in v2. Now we try to normalize naming to avoid running into problems with invalid names (symbols) for TypeScript types. 

What's not necessary correct is that perhaps we should preserve original filenames for files and just normalize names for types.

jchittoda:
I am also getting the same issue after upgraded to `typechain@2.0.0`. 

I had to upgrade to `typechain@2.0.0`, because the older version generates typing files with `BigNumber`. In our test cases we want to use `BN`. Hence, we had to upgrade it to `typechain@2.0.0`, which uses `BN`.

Can you suggest a better solution? So that we can use `BN` in generated typing files without upgrading to `typechain@2.0.0`

RobertoSnap:
Hi

Also running into this issue now.

PhABC:
Bump

This adds significant additional work since the names are different from the artifacts the typings are generated from.

PhABC:
Made a PR for it ; https://github.com/ethereum-ts/TypeChain/pull/290

quezak:
@krzkaczor what were the file/contract names resulting in invalid typescript symbols back when you originally changed this?

krzkaczor:
@quezak @PhABC there is a bunch of tests:
https://github.com/ethereum-ts/TypeChain/blob/master/packages/typechain/test/parser/normalizeName.test.ts

It looks like what @PhABC needs makes sense and was omitted by mistake. 

I will merge your (@PhABC) PR and add tests in a sec.

krzkaczor:
Take a look guys for the new tests and impl: https://github.com/ethereum-ts/TypeChain/pull/293/files

[...]

PR #291, opened 2020-10-28
assume type is address if type is unknown

KillerByte:
When given a abi definition with a type defined as such:

```
{
      "inputs": [
        {
          "internalType": "contract IConfigurableRightsPool",
          "name": "self",
          "type": "IConfigurableRightsPool"
        },
        {
          "internalType": "contract IBPool",
          "name": "bPool",
          "type": "IBPool"
        },
        {
          "internalType": "uint256",
          "name": "poolAmountIn",
          "type": "uint256"
        },
        {
          "internalType": "uint256[]",
          "name": "minAmountsOut",
          "type": "uint256[]"
        }
      ],
      "name": "exitPool",
      "outputs": [
        {
          "internalType": "uint256",
          "name": "exitFee",
          "type": "uint256"
        },
        {
          "internalType": "uint256",
          "name": "pAiAfterExitFee",
          "type": "uint256"
        },
        {
          "internalType": "uint256[]",
          "name": "actualAmountsOut",
          "type": "uint256[]"
        }
      ],
      "stateMutability": "view",
      "type": "function"
    },
    {
      "inputs": [
        {
          "internalType": "contract IConfigurableRightsPool",
          "name": "self",
          "type": "IConfigurableRightsPool"
        },
        {
          "internalType": "contract IBPool",
          "name": "bPool",
          "type": "IBPool"
        },
        {
          "internalType": "address",
          "name": "tokenOut",
          "type": "address"
        },
        {
          "internalType": "uint256",
          "name": "tokenAmountOut",
          "type": "uint256"
        },
        {
          "internalType": "uint256",
          "name": "maxPoolAmountIn",
          "type": "uint256"
        }
      ]
```

typechain fails to recognize the `IConfigurableRightsPool` type. I would not be so bothered by this normally, but when it fails, it stops the whole compilation.

instead of throwing an exception, this change will assume unknown types are addresses and return them as such.

this is a lot more graceful than killing the whole compile process because some libraries that I don't even use in TS had unknown types.

im not sure if there is a better way to approach this particular issue, but if I am missing something please suggest. For example, still return type address, but throw a warning instead of an error?


quezak:
@KillerByte can you please provide contract code of a minimal reproduction of this issue?

KillerByte:
here is the contract which my code fails on:

https://github.com/balancer-labs/configurable-rights-pool/blob/master/libraries/SmartPoolManager.sol

here is the buidler artifact input file:

https://gist.github.com/KillerByte/38ddb41e29295464541e0ece13cd8d3f

if I try building the buidler input file with:

```
typechain --target=ethers-v5 SmartPoolManager.json --show-stack-traces
```

I get output

```
ts-gen: Running TypeChain
ts-gen: artifacts/SmartPoolManager.json matched 1 files.
ts-gen: Processing artifacts/SmartPoolManager.json
Error occured:  Unknown type: IConfigurableRightsPool
Stack trace:  Error: Unknown type: IConfigurableRightsPool
    at Object.parseEvmType (node_modules/typechain/dist/parser/parseEvmType.js:53:11)
    at parseRawAbiParameterType (node_modules/typechain/dist/parser/abiParser.js:135:27)
    at parseRawAbiParameter (node_modules/typechain/dist/parser/abiParser.js:126:15)
    at Array.map (<anonymous>)
    at parseFunctionDeclaration (node_modules/typechain/dist/parser/abiParser.js:117:33)
    at node_modules/typechain/dist/parser/abiParser.js:28:28
    at Array.forEach (<anonymous>)
    at Object.parse (node_modules/typechain/dist/parser/abiParser.js:15:9)
    at Ethers.transformAbiOrFullJsonFile (node_modules/@typechain/ethers-v5/dist/index.js:53:38)
    at Ethers.transformFile (node_modules/@typechain/ethers-v5/dist/index.js:29:21)
```

KillerByte:
@quezak let me know if it is still necessary to isolate the problem further.

quezak:
@KillerByte I think your proposed solution is too broad -- we don't know what else the solidity compiler might put in the `type` field, so we may not want to apply this workaround _always_ :smiley: 

I'd update it to return `address` only if `type: "XXX"` and `internalType: "contract XXX"`, and if that is also false, throw an error.

@krzkaczor what do you think?

KillerByte:
@quezak part of my concern that I think we should explore in this PR is whether or not it is correct to throw an exception and die, or give a "best guess" with a warning, since when I use typechain with buidler, a single failure throws off the whole compilation process, and since I don't control these libraries, it causes problems.

quezak:
@KillerByte while this may be a good idea in some cases, always guessing `address` if we don't know the type should be can be _wrong_ in some cases that we're not yet aware of, and we don't want to generate files with invalid typings :smile:

IMHO we should apply [the fix I mentioned](https://github.com/ethereum-ts/TypeChain/pull/291#issuecomment-718957843) when we have enough hints the type is an address, but keep throwing an error if we _really don't know_ the type. An additional advantage is that someone will report it and we can fix the issue :smiley:  (and we try to accept simple PRs quickly, so this shouldn't be an issue for developement -- but if you're in a hurry, you can always use [patch-package](https://www.npmjs.com/package/patch-package))

If we really wanted to add some workarounds to not fail generation here, the proper way may be to return a special `type: "unknown"` placeholder, which would result in `any` being used for the property type -- doesn't stop generation, but also doesn't result in invalid typings. Waiting also for @krzkaczor to chip in with his opinion.

krzkaczor:
I basically agree with everything that @quezak said. We already have some similar heuristics in TC codebase. I hate it but what can you do 🤷 

To sum up:
1. If we cannot infer type but type: "XXX" and internalType: "contract XXX" then return `address`
2, If we still cannot infer type print out warning and return any type instead of interrupting whole process. 

@KillerByte let us know if you want to work on this. (2) seems like a breaking change for typechain core and all targets unfortunately so I would love to release it together with fix for: https://github.com/ethereum-ts/TypeChain/issues/234

KillerByte:
sure I will work on it. expect result within a few days.

KillerByte:
@krzkaczor let me know. not sure why circleci is complaining though.

quezak:
@KillerByte circleci complaints are right :smile: The generated `any` type is OK, but for this to work we also need to handle the newly generated type correctly in all the target generators. The CI output has all the paths that need fixing -- you can also check the same locally by running `yarn test`.

packages/typechain/src/parser/parseEvmType.ts
```diff
@@ -28,6 +31,9 @@ export type TupleType = { type: 'tuple'; components: EvmSymbol[]; originalType:
 // used only for output types
 export type VoidType = { type: 'void' }
 
+// used when type cannot be detected
+export type UnknownType = { type: 'any'; originalType: string }
```
krzkaczor, line 35: I would prefer to keep it consistent: IMO it should be:

```
export type UnknownType = { type: 'unknown'; originalType: string }
```

KillerByte:
@krzkaczor sorry for the commit spam, was having some computer issues. I'm concerned that additional tests may be necessary for the target libraries to verify that the use of `unknown` in the functions does not break TS compilation somehow. let me know if you agree that this is the case.

packages/target-ethers-v4/src/codegen/types.ts
```diff
@@ -40,6 +40,8 @@ export function generateInputType(evmType: EvmType): string {
       return 'string'
     case 'tuple':
       return generateTupleType(evmType, generateInputType)
+    case 'unknown':
```
krzkaczor, line 43: I think that `unknown` type should simply translate to `any`. 

As input parameter I think unknown doesn't make sense at all. Even for output I think it's just better to return `any` and do not thrive for 100% type correctness... but better DX, it's edge case anyway...  WDYT: @quezak

krzkaczor:
@KillerByte it looks good, I left 1 comm.

As for testing it, it would be best if you would add a contract that generates a need to return `unknown` evm type. This way, we can simply inspect generated output for the new file and treat it as a snapshot rather than real "test". Contracts used for tests are stored in `contracts` dir.

KillerByte:
> As for testing it, it would be best if you would add a contract that generates a need to return unknown evm type.

If our goal is to only include this "Unknown" type for ABI inputs/outputs that are not understood by the `parseEvmType` function, and our goal is for the `unknown` type to never need to be used, how do we create such a Contract? I bet we could do it if we could manufacture a fake ABI, but any such contracts would probably be slated for fix anyway.

krzkaczor:
Ahh sorry, you're right it shouldn't be possible to construct such a contract. In that case, I think this part is fine. I have a high level of certainty for the code generation part as it's typesafe and basically, compiler forced you to tweak types everywhere.

So two last things:
1) as mentioned in the previous comment I think that `unknown` should translate to `any`
2) please delete `package-lock.json`

packages/target-ethers-v4/src/codegen/types.ts
quezak: `unknown` would be more correct, but also would be annoying to use, let's go with `any` as the generated type

packages/typechain/src/parser/parseEvmType.ts
```diff
@@ -93,5 +99,18 @@ export function parseEvmType(rawType: string, components?: EvmSymbol[], internal
     return parseEvmType('uint8')
   }
 
-  throw new Error('Unknown type: ' + rawType)
+  if (internalType?.startsWith('contract')) {
+    return { type: 'address', originalType: rawType }
+  }
+
+  // unknown type
+  logger.warn(
```
krzkaczor, line 107: Can u use here es6 template strings to increase redability here?

packages/typechain/src/parser/parseEvmType.ts
```diff
@@ -93,5 +99,13 @@ export function parseEvmType(rawType: string, components?: EvmSymbol[], internal
...
   }
 
-  throw new Error('Unknown type: ' + rawType)
+  if (internalType?.startsWith('contract')) {
+    return { type: 'address', originalType: rawType }
+  }
+
+  // unknown type
+  logger.warn(`Could not parse type: ${rawType} with internal type: ${internalType}.\n\nPlease submit a GitHub Issue to the TypeChain team with the failing contract/library.`)
+  )
```
krzkaczor, line 108: FYI: this additional parenthesis breaks the CI

krzkaczor (approved):

krzkaczor:
Thanks for your hard work @KillerByte! This looks great! I want to merge a little bit more things before cutting a release so I'll let you know when this is released.

PR #294, opened 2020-11-03
Update "ethers-v5" example

rryter:
Hi there. 

I updated the `ethers-v5` example. Let me know what you would like changed, and I'll update it.

Cheers

### DIFF
--- a/examples/ethers-v5/package.json
+++ b/examples/ethers-v5/package.json
@@ -11,13 +11,13 @@
     "typecheck": "tsc --noEmit"
   },
   "devDependencies": {
-    "@typechain/ethers-v5": "^0.0.1",
+    "@typechain/ethers-v5": "^2.0.0",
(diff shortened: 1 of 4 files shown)


examples/ethers-v5/.prettierrc
```diff
@@ -0,0 +1 @@
+{ "printWidth": 100, "singleQuote": true }
```
krzkaczor, line 1: we already have `prettierrc` file in this project. If it's not picked up probably we need to explicitly point prettier to it. Reverting this should also avoid reforming index.ts

rryter: It's fixed.

I accidentally just opened the examples folder in my IDE, that's why it did not pick it up.

rryter:
I cleaned up the commits, should be fine now. 👍

quezak (approved):

PR #299, opened 2020-11-06
[ethers-v4-v5] move codegen contract factories to subdirectory

### LINKED ISSUE #225: Name collisions with `${contract.name}Factory.ts`
When using the ethers target with projects that have Factory contracts
- [ENSFactory.sol](https://github.com/aragon/aragonOS/blob/v4.4.0/contracts/factory/ENSFactory.sol)
- [ENS.sol](https://github.com/aragon/aragonOS/blob/c85d34e4bae0bf5b1ab78340b32e712d895179a7/contracts/lib/ens/ENS.sol)

The generated factory of ENS collides with the actual ENSFactory and overwrites its instance.

To solve this, maybe the name pattern could be customized? Or store instance files and factory files in different folders

0xGabi:
Fix #225 

The current PR generate factory contracts into a new subdirectory `factories`. The changes were included in ea36162fe1e193a4484ee0511fff2847fefc7c0d and new types generated in a subsequent commit.

Let me know if the changes are align with what you have in mind. Also if you have some guidelines on what we should tests here. Thanks


krzkaczor (commented):
This looks very good!

Can you delete old factories? Maybe `rm -rf` whole types dir locally and then rerun tests? All file generation should be deterministic so it should go fine.

Also, I don't see regenerated updated `index.ts` 🤔

0xGabi:
I did delete the old factories, but I think GitHub detected it as a file rename plus the changes.

About the `index.ts` you are right. Having a closer look on why is not writing the file 👍

quezak (approved):
Changes looks great! I think the additional `PayableFactory` contract is sufficient for testing this change.

If you can, please also update:
* example in [ethers v4 readme](https://github.com/ethereum-ts/TypeChain/blob/master/packages/target-ethers-v4/README.md)
* example in [ethers v5 readme](https://github.com/ethereum-ts/TypeChain/blob/master/packages/target-ethers-v5/README.md)
* [ethers v4 example code](https://github.com/ethereum-ts/TypeChain/blob/master/examples/ethers-v4/src/index.ts)
* [ethers v5 example code](https://github.com/ethereum-ts/TypeChain/blob/master/examples/ethers-v5/src/index.ts)
* doesn't look like truffle examples use the factories at all

krzkaczor:
@0xGabi you're right it's fine. 

Don't worry about examples I will update them after releasing the package as I think they don't use local packages anyway.

I think the last remaining thing is the lack of changes for index files for ethers target. I can take a closer look if you're having problems with this... It's strange because I would expect that CI would fail as it would detect some changes...

krzkaczor:
Changes generated by this exact line should be somewhere in the diff but I think they are missing 🤔 

https://github.com/ethereum-ts/TypeChain/pull/299/files#diff-bd01eaab8bcc9d57d458cd18a611ffe2f79650335f75d4b577db538a2be7ecceR131

0xGabi:
Thanks for updating the packages. About the `index.ts` I'm in the process of debugging the generation to have further details. And I agree that is strange that the diff didn't pick the changes

0xGabi:
I can confirm that after linking the target package with the changes on a repo I was working, the types and even the `index.ts` generates fine 🤔

Still trying to understand why the example write an empty string though.

krzkaczor:
I think that maybe local packages are not linked correctly after the recent version bumping. You can leave it and I will take a closer look at it tomorrow - it's clearly out of scope of the original issue. I think, the gist of the problem is that I couldn't make  `yarn`  to always link local packages and perhaps it's trying to use version from the npm.

0xGabi:
Sound good, thanks for let me know!

krzkaczor:
Merged on tmp branch.

PR #300, opened 2020-11-08
Resolve name collisions with factory contracts

krzkaczor:
Closes: https://github.com/ethereum-ts/TypeChain/issues/225

Done on top of @0xGabi work. So feel free only to check last commit.

It fixes two unrelated issues:

1) We didn't see changes in generated `index.ts` before, because TC was not generating reexports for abstract factories and our current tests rely only on abstract factories. Now all factories are exported. **Here I made the assumption that all contracts will have factories generated (normal or abstract). Please verify that @quezak @zemse.**

2) Reexported symbols can still have conflicting names. Reexporting was redesigned to first generate types for user-provided contracts and later for TC factories but ensuring that extracted symbols are unique.

Let me know what do you think and I will migrate this changes to `ethers-v4` target as well.

Sidenote: various factories seems to not be tested properly right now :/ hopefully next week I will improve it.


zemse:
> Here I made the assumption that all contracts will have factories generated (normal or abstract).

This should be fine since an abstract factory file is generated with `Factory` append, same like in the case of normal contracts.

I was thinking about how the `_` would be during developer experience.

Consider two contracts:

```
contract Charity {}
contract CharityFactory {}
```

The first one will be processed before the other.

```ts
import { Charity, CharityFactory, _CharityFactory, __CharityFactory } from '...'
```

It's difficult to predict what is what at a first glance, would need to have a look at docs how this work.

If this is going to be a breaking change, I just have a suggestion that we can reconsider the initial approach of appending `Factory`. Given that in solidity contract naming, camel case is followed, maybe we can append `_Factory` or `__Factory` or `_factory` or `__factory` instead (inspired from  https://github.com/ethereum-ts/TypeChain/issues/225#issuecomment-723147526). In that case:

```ts
import { Charity, Charity__factory, CharityFactory, CharityFactory__factory } from '...'
```

The above might look weird but at least it is clear what is what. And just in case someone named their contract as `Charity__factory`, then to prevent typechain's index file name collision, the approach in this PR takes care of it. Even the files can be renamed by the same naming: 

```
Charity.ts
Charity__factory.ts
CharityFactory.ts
CharityFactory__factory.ts
```

I understand that this opinion is late to the discussion, after @0xGabi's and your hard work. So if it's not possible to be considered, present ones may be moved forward.

quezak:
I agree with @zemse. If we're releasing a breaking change anyway, let's make the names more developer friendly by making them not colliding at all.

Downsides to the current proposal in my opinion:
* the factory name is insonsistent in different scenarios
* the actual factory class name is still colliding with the other contract name
* the factory class is imported as `_ContractFactory`, but in runtime its name is still `ContractFactory` which can be confusing
* the prefix in `_ContractFactory` is not really meaningful, or even counter-intuitive, because an underscore prefix is most commonly used to indicate something unimportant, internal or "kinda private"

quezak:
> Here I made the assumption that all contracts will have factories generated (normal or abstract).

Yes, contracts with bytecode get a regular factory, and in `afterRun` all other contracts get an abstract factory: https://github.com/ethereum-ts/TypeChain/blob/b5aac4164c72c90bbea051fe13bf5e60e614da5e/packages/target-ethers-v4/src/index.ts#L110-L112

krzkaczor:
My initial thinking was to avoid bigger breaking changes like changing the naming scheme but I agree that this might be a good opportunity to do so. Thanks for comments guys. 

I will go with `Contract__factory` pattern. This will also mean that factory classes will have to be renamed to match this pattern...

krzkaczor:
Okay it's done. Please re-review @zemse @quezak .

zemse (approved):
LGTM! Thank you!

packages/target-ethers-v4/src/common.ts
```diff
@@ -0,0 +1 @@
+export const FACTORY_POSTFIX = '__factory'
```
quezak, line 1: it's it more commonly named `suffix`? :D

quezak (approved):
Looks good! Please check if ether targets readme or examples directory need updating.

0xGabi:
Thanks a lot guys!! 🙌

Issue #306, opened 2020-11-20
Packages shouldn't contain tests


krzkaczor:
`package.json` should whitelist files with `files`. Now tests are shipped to npm as part of the package 🤦

PR #308, opened 2020-11-30
Migrate CI to Github Actions

krzkaczor:
Closes: https://github.com/ethereum-ts/TypeChain/issues/306


.changeset/purple-geese-listen.md
```diff
@@ -0,0 +1,5 @@
+---
```
quezak, line 1: why does this file have such a weird name?

krzkaczor: 😆 everyone asks me this question when I introduce changesets to a new project. 

Apparently, this is how changesets works, it seems like it creates files with random names to avoid conflicts during merges. It's an odd choice if you ask me because it could be something less weird like commit SHA but I don't know/don't really care. CC: @Andarist

quezak: yeah, it would be much clearer if it was something like `changeset-${timestamp}-${commit hash first 8 digits}`

Andarist: Commit hash is not really an option because we can't really put a hash into it until a commit is created, right? 😉 The only possibility would be to put an older commit hash in the filename but then this would be IMHO even more confusing.

Overall - the filename doesn't matter. You can freely create changeset files by hand. Using timestamps also has some inconsistency problems associated with it - for example, the bot posts a link like this: "Click here if you're a maintainer who wants to add another changeset to this PR". If you click on it you will be redirected to a Github page on which you will be creating a new changeset file with the filename already prefilled and this can't be dynamic, we have to put the filename there upfront.

quezak: @Andarist thanks for the explanation. If the filename doesn't matter I think it's yet another reason to give it a name that's _obviously and intentionally random_, not one which makes people wonder like here :smile: especially people who are not fully familiar with changeset, like Chris mentioned it's quite common. Maybe something like `changeset-{uuid}`?

PR #311, opened 2020-12-02
Improve types for tuples

krzkaczor:
Closes: https://github.com/ethereum-ts/TypeChain/issues/232

### DIFF
--- a/.changeset/fair-paws-pay.md
+++ b/.changeset/fair-paws-pay.md
@@ -25,4 +25,3 @@ x(overrides?: CallOverrides): Promise<[BigNumber]>;
 ```
 
 The difference is that now you can use standard array destructuring while working with output types.
-There should be no changes needed to existing code.
--- a/.changeset/fair-paws-pay.md
+++ b/.changeset/fair-paws-pay.md
@@ -1,0 +1,28 @@
+---
+'@typechain/ethers-v4': major
+'@typechain/ethers-v5': major
+---
+
+Changed return type of functions from a object with number indexes, to an array merged with object containing named
+outputs.
+
+Before, solidity function like this:
+
+```
+function x() public pure returns (uint256)
+```
+
+Generated such method signature:
+
+```typescript
+x(overrides?: CallOverrides): Promise<{0: BigNumber}>;
+```
+
+New output is:
+
+```typescript
+x(overrides?: CallOverrides): Promise<[BigNumber]>;
+```
+
+The difference is that now you can use standard array destructuring while working with output types.
+There should be no changes needed to existing code.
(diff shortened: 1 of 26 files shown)


.changeset/fair-paws-pay.md
```diff
@@ -0,0 +1,27 @@
...
+x(overrides?: CallOverrides): Promise<{0: BigNumber}>;
+```
+
+New output is:
+
+```typescript
+x(overrides?: CallOverrides): Promise<[BigNumber]>;
+```
+
+The difference is that now you can standard array destruction while working with output types.
```
quezak, line 27: image
([docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment))
```suggestion
The difference is that now you can use standard array destructuring while working with output types.
There should be no changes needed to existing code.
```

packages/target-ethers-v5/src/codegen/types.ts
```diff
@@ -82,11 +79,15 @@ export function generateTupleType(tuple: TupleType, generator: (evmType: EvmType
...
 
-export function generateOutputTupleType(tuple: TupleType) {
-  return (
-    '{' +
-    tuple.components.map((component) => `${component.name}: ${generateOutputType(component.type)} ,`).join('\n') +
-    tuple.components.map((component, index) => `${index}: ${generateOutputType(component.type)}`).join(', ') +
-    '}'
-  )
+/**
+ * always return an array type if there are named outputs, attach them to that type
```
quezak, line 83:
```suggestion
 * always return an array type; if there are named outputs, merge them to that type
```

packages/target-truffle-v5-test/types/truffle-contracts/DataTypesInput.d.ts
```diff
@@ -67,7 +67,7 @@ export interface DataTypesInputInstance extends Truffle.ContractInstance {
     input1: number | BN | string,
     input2: number | BN | string,
     txDetails?: Truffle.TransactionDetails
-  ): Promise<[BN, BN]>;
+  ): Promise<{ 0: BN; 1: BN }>;
```
quezak, line 70: funny that for truffle the change is excatly the opposite of the ethers change :sweat_smile:

quezak (approved):

packages/target-truffle-v5-test/types/truffle-contracts/DataTypesInput.d.ts
krzkaczor: Yup :D but anyway, now types correspond to real runtime values so it's fine.

PR #317, opened 2021-01-05
Issues/315 truffle-v5: Some properties of `Truffle.ContractInstance` are missing

### LINKED ISSUE #315: truffle-v5: Some properties of `Truffle.ContractInstance` are missing
For example, properties of `Overloads` of `packages/target-truffle-v5-test/test/Overloads.test.ts` are different between type and runtime.

type:

runtime:

missing properties:
- abi
- allEvents
- ~getPastEvents~
- send
- sendTransaction

odanado:
Close #315

### DIFF
--- a/packages/target-truffle-v5/package.json
+++ b/packages/target-truffle-v5/package.json
@@ -48,7 +48,6 @@
     "typescript": "3.9.0-beta",
     "web3": "^1.2.6",
     "web3-core": "^1.3.1",
-    "web3-eth": "^1.3.1",
     "web3-utils": "^1.3.1"
   }
 }
--- a/packages/target-truffle-v5/package.json
+++ b/packages/target-truffle-v5/package.json
@@ -30,8 +30,8 @@
     "test:fix": "yarn lint:fix && yarn format:fix && yarn test && yarn typecheck"
   },
   "peerDependencies": {
-    "typechain": "^4.0.0",
-    "truffle": "^5.0.0"
+    "truffle": "^5.0.0",
+    "typechain": "^4.0.0"
   },
   "devDependencies": {
     "@types/mocha": "^5.2.7",
@@ -44,8 +44,11 @@
     "ganache-cli": "^6.9.1",
     "mocha": "^8.2.0",
     "ts-node": "^8.5.4",
+    "typechain": "*",
     "typescript": "3.9.0-beta",
     "web3": "^1.2.6",
-    "typechain": "*"
+    "web3-core": "^1.3.1",
+    "web3-eth": "^1.3.1",
+    "web3-utils": "^1.3.1"
   }
 }
--- a/packages/target-truffle-v5/package.json
+++ b/packages/target-truffle-v5/package.json
@@ -30,8 +30,8 @@
     "test:fix": "yarn lint:fix && yarn format:fix && yarn test && yarn typecheck"
   },
   "peerDependencies": {
-    "truffle": "^5.0.0",
-    "typechain": "^4.0.0"
+    "typechain": "^4.0.0",
+    "truffle": "^5.0.0"
   },
   "devDependencies": {
     "@types/mocha": "^5.2.7",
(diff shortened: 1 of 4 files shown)


packages/target-truffle-v5/static/types.d.ts
```diff
@@ -78,10 +84,20 @@ declare namespace Truffle {
...
+    topics?: string[];
+  }
+
   interface ContractInstance {
     address: string
     contract: any
     transactionHash: string
+    abi: AbiItem[]
+    allEvents(params?: EventOptions): EventEmitter
+    send(value: Required<TransactionConfig>["value"], txParams?: TransactionConfig): PromiEvent<TransactionReceipt>
```
odanado, line 99: ref: https://github.com/trufflesuite/truffle/blob/2b247527f3d3c402f98ae1a968994e3ffcbf56be/packages/contract/lib/contract/index.js#L129

packages/target-truffle-v5/static/types.d.ts
```diff
@@ -78,10 +84,20 @@ declare namespace Truffle {
...
+    fromBlock?: BlockType;
+    topics?: string[];
+  }
+
   interface ContractInstance {
     address: string
     contract: any
     transactionHash: string
+    abi: AbiItem[]
+    allEvents(params?: EventOptions): EventEmitter
```
odanado, line 98: ref: https://github.com/trufflesuite/truffle/blob/6f937e8e3f03499751d7ede0e05d5f8c19ac271a/packages/contract/lib/execute.js#L376
ref: https://web3js.readthedocs.io/en/v1.2.0/web3-eth-contract.html?highlight=getPastEvents#events-allevents

packages/target-truffle-v5/static/types.d.ts
```diff
@@ -78,10 +84,20 @@ declare namespace Truffle {
...
+    send(value: Required<TransactionConfig>["value"], txParams?: TransactionConfig): PromiEvent<TransactionReceipt>
+    sendTransaction(transactionConfig: TransactionConfig): PromiEvent<TransactionReceipt>
```
odanado, line 100: > This is promisified like all available contract instance functions, and has the same API as web3.eth.sendTransaction but without the callback.
https://www.trufflesuite.com/docs/truffle/getting-started/interacting-with-your-contracts

packages/target-truffle-v5/package.json
```diff
@@ -30,8 +30,8 @@
     "test:fix": "yarn lint:fix && yarn format:fix && yarn test && yarn typecheck"
   },
   "peerDependencies": {
-    "typechain": "^4.0.0",
```
krzkaczor, line 33: Can we revert this change as it's just reordering?

packages/target-truffle-v5/package.json
```diff
@@ -44,8 +44,10 @@
     "ganache-cli": "^6.9.1",
     "mocha": "^8.2.0",
     "ts-node": "^8.5.4",
+    "typechain": "*",
     "typescript": "3.9.0-beta",
     "web3": "^1.2.6",
-    "typechain": "*"
+    "web3-core": "^1.3.1",
```
krzkaczor, line 50: why do we need these new deps?

krzkaczor (changes requested):
Tests on CI fail because you need to commit new typing files (we use it as a snapshot). 

To do so follow run `yarn test:fix` locally and commit changed files in: `packages/target-truffle-v5-test/types/truffle-contract`

packages/target-truffle-v5/package.json
odanado: Yes. I can revert.
But, `package.json` has been reordered by `yarn`. So, the diff may occur again.

packages/target-truffle-v5/package.json
odanado: Because we imported type of `web3-utils` and `web3-core` in `packages/target-truffle-v5/static/types.d.ts`.
https://github.com/ethereum-ts/TypeChain/pull/317/files/5e749a452ab9e602e8f9e773d114c44dddbc0f10#diff-6f45ea93d39575567adc3eb0b242f1ec2a2eda912f9436c5d883111c02fced66

PR #320, opened 2021-01-25
Fix signature of overloaded methods with struct in arguments

### LINKED ISSUE #295: Typechain & ethers - Wrong Function Fragments
The generated types for the function fragments of a contract object are wrong. As soon as an argument is of type ``tuple`` is the function fragment name/signature not correct. As I have seen during a quick check that TypeChain does not parse the whole ``tuple`` it just uses the type ``tuple``.. means: ``myFunction(tuple)`` instead of e.g.: ``myFunction((uint, address))``. The same behavior is for event fragments.

nicklatkovich:
Closes #295

### DIFF
--- a/packages/typechain/src/utils/signatures.ts
+++ b/packages/typechain/src/utils/signatures.ts
@@ -16,9 +16,9 @@ export function getIndexedSignatureForEvent(event: EventDeclaration): string {
 
 export function getArgumentForSignature(argument: AbiParameter): string {
   if (argument.type.originalType !== 'tuple') return argument.type.originalType
-  return `(${(argument.type as TupleType).components.map((i) => getArgumentSignature(i)).join(',')})`
+  return `(${(argument.type as TupleType).components.map((i) => getArgumentForSignature(i)).join(',')})`
 }
 
 export function getSignatureForFn(fn: FunctionDeclaration): string {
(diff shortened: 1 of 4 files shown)


packages/typechain/src/utils/signatures.ts
```diff
@@ -13,6 +14,11 @@ export function getIndexedSignatureForEvent(event: EventDeclaration): string {
   return `${event.name}(${indexedType.map((e) => e.type.originalType).join(',')})`
 }
 
+export function getArgumentSignature(argument: AbiParameter): string {
```
quezak, line 17: just a naming nitpick -- the whole string returned by `getSignatureForFn` is a function signature, a single input type is not called a signature :)
```suggestion
export function getArgumentForSignature(argument: AbiParameter): string {
```

quezak (approved):
Looks good 🆒 just one nitpick to fix.

Thanks for the contribution!

quezak:
@nicklatkovich @krzkaczor what impact will this have on existing generated code? Will correct code need updates or should it be fine?

krzkaczor:
This looks great! Thanks for the fix @nicklatkovich. It looks like a patch change for me as without it some existing signatures were simply wrong and couldn't be used.

krzkaczor:
Released in: https://github.com/ethereum-ts/TypeChain/releases/tag/typechain%404.0.2

PR #321, opened 2021-01-29
[ethers-v5] Add events support

### LINKED ISSUE #249: Adding event type for ethers
Hello!

Thanks for the great project! 

It seems that typechain does not output the event type when it is generated by specifying ethers-v4 as the target.
This is inconvenient because I don't know the type when implementing the event handler.

**Sample Code**
```
pragma solidity ^0.6.4;
pragma experimental ABIEncoderV2;
contract Test {
  event TestEvent(address sender, string message);
  function test() public {
    emit TestEvent(msg.sender, 'test Message!');
  }
}
```

**Output Result**
```Test.d.ts
interface TestInterface extends Interface {
  events: {
    TestEvent: TypedEventDescription<{
      encodeTopics([sender, message]: [null, null]): string[];  **// type is null**
    }>;
  };
}

export class Test extends Contract {
  filters: {
    TestEvent(sender: null, message: null): EventFilter;  **// type is null**
  };
}
```

**Expected Behavior**
It would be helpful if you could output something like the following
export type TestEvent = [string, string]

If struct type is used...
export type TestEvent = [string, { id: string, name: string }]

### LINKED ISSUE #303: Add type support for ethers v5 queryFilter
```solidity
event Transfer(address sender, address receiver, uint256 value);
```

```ts
const filter = contractInst.filters.Transfer(); // TransferFilter

const logs = await contractInst.queryFilter(filter) // Overloaded for each filter

logs[0].args // before `any`, but after this feature it should be `{sender: string, receiver: string, value: BigNumber}`
```

I'll give this a shot while doing my next project starting probably end of this month or next month.

zemse:
This PR adds the missing events support to the ethers v5 target and intends to close #303 and close #249. Thanks @quezak for https://github.com/ethereum-ts/TypeChain/issues/303#issuecomment-729030874

## Usage

```ts
const filter = contract.filters.Transfer(); // TypedEventFilter<>
const result = await contract.queryFilter(filter); // TypedEvent<>

result[0].args.from // type support for named event parameters
result[0].args[0] // type support by index

contract.on(filter, (from, to, value, event) => {
  from; // string
  to; // string
  value; // BigNumber
  event; // TypedEvent<>
})
```

## To test this in your project

1. `git clone -b 303-ethers-v5-add-event-support https://github.com/zemse/TypeChain.git`
2. `cd TypeChain`
3. `yarn` (you need yarn installed if you don't)
4. `yarn build` (this generates dist dir which is imp)
5. Link this custom target to your project by placing the appropriate path to the target: `typechain --target ../TypeChain/packages/target-ethers-v5/` ([Docs](https://github.com/ethereum-ts/TypeChain#usage))

Just in case if that sounds too much, I have temporarily published it to npm for testing:

1. `npm install typechain-target-ethers-v5 --save-dev` (later you can just replace the package with official one)
2. `typechain --target ethers-v5` ([Docs](https://github.com/ethereum-ts/TypeChain#usage))


zemse:
Hi @krzkaczor, there is the following error when running the `typecheck` script (failing in CI) but my typescript compiler (v4.1.3) does not complain about it and seems to work fine. Can you help how this issue can be resolved?

```
types/commons.ts(15,20): error TS1256: A rest element must be last in a tuple type.
error Command failed with exit code 1.
```

packages/target-ethers-v5/src/codegen/index.ts
```diff
@@ -64,11 +65,25 @@ export function codegenContractTypings(contract: Contract) {
...
+    once(eventName: string, listener: Listener): this;
+    removeListener(eventName: string, listener: Listener): this;
+    removeAllListeners(eventName?: string): this;
+
+    listeners<T, G>(eventFilter?: TypedEventFilter<T, G>): Array<TypedListener<T, G>>;
+    off<T, G>(eventFilter: TypedEventFilter<T, G>, listener: TypedListener<T, G>): this;
+    on<T, G>(eventFilter: TypedEventFilter<T, G>, listener: TypedListener<T, G>): this;
+    once<T, G>(eventFilter: TypedEventFilter<T, G>, listener: TypedListener<T, G>): this;
+    removeListener<T, G>(eventFilter: TypedEventFilter<T, G>, listener: TypedListener<T, G>): this;
+    removeAllListeners<T, G>(eventFilter: TypedEventFilter<T, G>): this;
```
quezak, line 80: I think the more specific overloads should be first, since TS checks them top-to-bottom IIRC

packages/target-ethers-v5/src/index.ts
```diff
@@ -129,6 +133,20 @@ export default class Ethers extends TsGeneratorPlugin {
     return allFiles
   }
 
+  private genCommons(): string {
+    return `
+    import { EventFilter, Event } from 'ethers'
+    import { Result } from '@ethersproject/abi'
+
+    export interface TypedEventFilter<T, G> extends EventFilter {}
```
quezak, line 141: why are the `T` & `G` args here if they're not used?

packages/target-ethers-v5/src/index.ts
```diff
@@ -129,6 +133,20 @@ export default class Ethers extends TsGeneratorPlugin {
...
+
+    export interface TypedEvent<T extends Result> extends Event {
+      args: T;
+    }
+    
+    export type TypedListener<T extends Array<any>, G> = (...listenerArg: [...T, TypedEventFilter<T, G>]) => void;`
```
quezak, line 147: Can we use whole word type names instead of just `T` & `G`? I didn't use ethers events for some time so now I don't know what are these args without refering to ethers docs (and frankly the docs don't help much either 😅)

quezak (commented):
Some initial comments before I try this out on my contracts.

Random thought: is it a good idea to skip the whole typed listeners section if a contract has no defined events?

packages/target-ethers-v5/src/index.ts
zemse: The `event.args` in ethers js allows access as a array or by parameter name. For e.g. `Transfer(address,address,uint256)` the queryFilter will output the following interfce:

```ts
import { Event } from 'ethers'

interface TransferEvent extends Event {
  args: [string, string, BigNumber] & { from: string, to: string, value: BigNumber }
}
```

So there are two components `[string, string, BigNumber]` and `{ from: string, to: string, value: BigNumber }`. Initially I was not separating them out, but the `contract.on` needs the first component to be able to have types on the callback.

I agree they should be renamed instead of leaving T & G. What do you think about `T` =>`ArrayComponent` and `G` => `ObjectComponent` or any better ideas?

packages/target-ethers-v5/src/index.ts
zemse: Actually the `T` & `G` contain type info utilised by `queryFilter` and other methods for output

packages/target-ethers-v5/src/codegen/index.ts
zemse: > is it a good idea to skip the whole typed listeners section if a contract has no defined events?

It's a definitely great idea. Since even if these typed listener section lies there useless for contracts without events, I don't think it would affect normal functionality (lmk if it does). Would it be fine if we do that in a separate PR? If not I can do that here.

krzkaczor:
@zemse as for this TS error, it seems that it works in TS 4.0+.

You should get the same error if you use `yarn.lock` 🤔 Anyway, it doesn't matter, we can bump TS to 4.0 (or even newer?). And bump peer dependency from "typescript": ">=3.8.0",` to 4+ in `packages/target-ethers-v5/package.json`.


Soo, could you:
1. Normalize all package.json dependencies on typescript to 4.0
2. Bump peer dependency? in `packages/target-ethers-v5/package.json`

I will deploy it as major version of ethers-v5 to avoid breaking something.

Thanks!

krzkaczor:
@zemse can we add a test (can be the same thing as you put in this PR description) proofing that this works?

zemse:
I've added some tests pls review

krzkaczor:
@zemse thanks for this! Released in: https://github.com/ethereum-ts/TypeChain/releases/tag/%40typechain%2Fethers-v5%406.0.0

fubar:
@zemse how is this supposed to work? Maybe I'm missing something but the first line in your example `contract.filters.Transfer()` is invalid because the declaration of that function is generated with arguments:

```ts
Transfer(
  from: string | null,
  to: string | null,
  value: null
): TypedEventFilter<
  [string, string, BigNumber],
  { from: string; to: string; value: BigNumber }
>;
```

Which means those 3 arguments are required.

zemse:
Oh right, you need to do the following, will update the description too.

```ts
contract.filters.Transfer(null, null, null)
```

As a side note: Actually ethers.js on it's core is fine without passing any args, so the args should be made optional in the target too. Will make a PR soon.

fubar:
@zemse sounds good, thanks!

PR #327, opened 2021-02-23
[web3] accept BN as input for numbers

DanielZlotin:
web3 can handle BN values for number inputs.

This PR adds ` | BN` to all number inputs, making working with web3 + TypeScript easier.
Instead of `contract.methods.foo(bn.toString())`
we can now just `contract.methods.foo(bn)`.

There is a runtime error when passing BN as properties of structs or arrays:
```
     Error: invalid number value (arg="input1", coderType="uint8", value="1")
      at Object.throwError (/TypeChain/node_modules/ethers/utils/errors.js:68:17)
      at CoderNumber.encode (/TypeChain/node_modules/ethers/utils/abi-coder.js:353:20)
      at /TypeChain/node_modules/ethers/utils/abi-coder.js:605:59
      at Array.forEach (<anonymous>)
...
```
This appears to be a bug in `web3-eth-contract`'s `abi-coder`, which under the hood, [uses `ethers`'s `abi-coder`](https://github.com/ChainSafe/web3.js/blob/1.x/packages/web3-eth-abi/src/index.js#L28) (wut 😆 ),
IMHO can be ignored here, as its not under this repo's responsibility.

see similar issue https://github.com/ChainSafe/web3.js/issues/3795

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
-.idea/
 node_modules
 dist
 .nyc_output
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+.idea/
 node_modules
 dist
 .nyc_output
--- a/packages/target-web3-v1-test/test/DataTypesInput.test.ts
+++ b/packages/target-web3-v1-test/test/DataTypesInput.test.ts
@@ -3,6 +3,7 @@ import { typedAssert, q18 } from 'test-utils'
 
 import { createNewBlockchain, deployContract } from './common'
 import { DataTypesInput } from '../types/DataTypesInput'
+import BN from 'bn.js'
 
 describe('DataTypesInput', () => {
   let contract!: DataTypesInput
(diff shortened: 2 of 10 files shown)


quezak:
> There is a runtime error when passing BN as properties of structs or arrays

Maybe it's not a problem with structs or arrays, but with smaller integers in general? Maybe returning just `number` when the  type is just `int8/uint8/enum` would solve/mitigate this issue? We did a similar thing for [output types in ethers target](https://github.com/ethereum-ts/TypeChain/blob/master/packages/target-ethers-v5/src/codegen/types.ts#L53) (since ethers actually return a plain number for small integers)

quezak:
Update: ☝🏻 no, it's not -- the PR adds a test for `input_uint8` where a BN instance is passed and it works.

It's also not an issue that the struct fields should are uint8, since they [in fact are uint256](https://github.com/ethereum-ts/TypeChain/blob/8b697e9f0058a40942240c8580ef27fcc6d9da84/contracts/DataTypesInput.sol#L40). Maybe web3-eth-contract incorrectly integrates the abi coder, because the error you posted says `coderType="uint8"`?

@DanielZlotin would you create an issue in `web3-eth-contract` about this, so we can reference it here and the underlying problem is not lost? Then we probably shouldn't have issues merging this PR as is :)

DanielZlotin:
https://github.com/ChainSafe/web3.js/issues/3920

packages/target-web3-v1-test/test/DataTypesInput.test.ts
```diff
@@ -42,14 +49,23 @@ describe('DataTypesInput', () => {
 
     typedAssert(await contract.methods.input_stat_array(['1', '2', '3']).call(), ['1', '2', '3'])
     typedAssert(await contract.methods.input_stat_array([1, 2, 3]).call(), ['1', '2', '3'])
+    // TODO this fails due to an issue in web3 abi coder handling of inner BN:
```
quezak, line 52: Let's link the web3 issue here and I think we can merge this, @krzkaczor what do you think?

quezak (approved):

quezak:
> [ChainSafe/web3.js#3920](https://github.com/ChainSafe/web3.js/issues/3920)

Thanks, great job narrowing down the reproduction!

.gitignore
```diff
@@ -1,3 +1,4 @@
+.idea/
```
quezak, line 1: You can add `.idea` to your [global gitignore](https://stackoverflow.com/questions/7335420/global-git-ignore) -- we don't want to end up with a gitignore for over 9000 IDEs and tools here :)
```suggestion
```

krzkaczor:
Thanks @DanielZlotin! Released as https://github.com/ethereum-ts/TypeChain/releases/tag/%40typechain%2Fweb3-v1%402.2.0

PR #331, opened 2021-02-26
Fix ethers-v5 code-generation for no-argument events

dwardu:


quezak (approved):
LGTM, thanks!
Does this fix https://github.com/ethereum-ts/TypeChain/pull/321#issuecomment-785557015 ?

krzkaczor:
CC: @zemse

dwardu:
@quezak The code introduced in #321 generates bad code for events that have 0 arguments. I think the comment you pointed to is different, because the `Transfer` event being discussed there has 3 arguments.

The way the “output type” is being generated is by calling `generateOutputTypes` (which calls `generateOutputComplexType`) , which returns “complex output types” like `'[boolean, BigNumber] & {value1: boolean,value2: BigNumber }'`… and then taking this string and replacing `&` with `,` to produce the type-parametrization necessary for `TypedEventFilter<EventArgsArray, EventArgsObject>`. It’s a bit hacky, and the hack doesn't work for 0-arg events, because internally `generateOutputComplexType` [does not generate the 2nd argument for 0-arg events](https://github.com/ethereum-ts/TypeChain/blob/e4ccc3d4ab3abfe015b431ada30132e47d921794/packages/target-ethers-v5/src/codegen/types.ts#L89).

My fix is to, for the 0-arg event case, avoid the “hack” and use `'[], {}'` as the “output type” directly.

quezak:
> @quezak The code introduced in #321 generates bad code for events that have 0 arguments. I think the comment you pointed to is different, because the `Transfer` event being discussed there has 3 arguments.
> 
> The way the “output type” is being generated is by calling `generateOutputTypes` (which calls `generateOutputComplexType`) , which returns “complex output types” like `'[boolean, BigNumber] & {value1: boolean,value2: BigNumber }'`… and then taking this string and replacing `&` with `,` to produce the type-parametrization necessary for `TypedEventFilter<EventArgsArray, EventArgsObject>`. It’s a bit hacky, and the hack doesn't work for 0-arg events, because internally `generateOutputComplexType` [generates the 2nd output-type only for 1+ outputs](https://github.com/ethereum-ts/TypeChain/blob/e4ccc3d4ab3abfe015b431ada30132e47d921794/packages/target-ethers-v5/src/codegen/types.ts#L89).
> 
> My fix is to, for the 0-arg event case, avoid the “hack” and use `'[], {}'` as the “output type” directly.

OK, thanks for the explanation -- I wanted to confirm if this solves that linked case too.
Thanks for the PR!

dwardu:
You’re welcome @quezak. Thank _you_. TypeChain is a great tool, and I am happy to contribute this tiny fix.

krzkaczor:
Thanks @dwardu! This looks good. I just want to hear @zemse's opinion on this before merging. 

I will make sure to release this ASAP as it seems like an urgent fix.

zemse:
Thanks @dwardu for the fix! I regret that this error was introduced due to the previous PR and sorry for the inconvenience in your live projects. Since the [`generateOutputTypes`](https://github.com/ethereum-ts/TypeChain/blob/e4ccc3d4ab3abfe015b431ada30132e47d921794/packages/target-ethers-v5/src/codegen/index.ts#L278) generates the 2nd output-type [only for 1+ outputs](https://github.com/ethereum-ts/TypeChain/blob/e4ccc3d4ab3abfe015b431ada30132e47d921794/packages/target-ethers-v5/src/codegen/types.ts#L89), the case for 0 outputs was needed to be handled, which this PR takes care of. 👍

dwardu:
Hi @zemse, no problem at all!, and thanks for for programming the events support in the first place.

krzkaczor:
Thanks for the fix. Released as: https://github.com/ethereum-ts/TypeChain/releases/tag/%40typechain/ethers-v5%406.0.1

dwardu:
That’s great. Thanks :tada:

PR #334, opened 2021-03-03
Update ethers-v5 example

tuler:
Updates dependencies, and the new factory class name.


krzkaczor:
Thanks @tuler! This looks good. I will get this merged tmrw.

PR #337, opened 2021-03-06
Fix unused type parameters ts error

### LINKED ISSUE #335: Unused type parameters in typechain/ethers-v5 6.0.1
After bumping `@typechain/ethers-v5` from `5.0.5` to `6.0.1` my project stops compiling when `noUnusedParameters` compiler option is enabled in TS 4.0.1.

The TypeChain-generated file `commons.ts` triggers an error:
```ts
contract-bindings/ethers/commons.ts:8:34 - error TS6205: All type parameters are unused.

8 export interface TypedEventFilter<EventArgsArray, EventArgsObject>
```
This is its body:
```ts
/* Autogenerated file. Do not edit manually. */
/* tslint:disable */
/* eslint-disable */

import { EventFilter, Event } from "ethers";
import { Result } from "@ethersproject/abi";

export interface TypedEventFilter<EventArgsArray, EventArgsObject>
  extends EventFilter {}

export interface TypedEvent<EventArgs extends Result> extends Event {
  args: EventArgs;
}

export type TypedListener<
  EventArgsArray extends Array<any>,
  EventArgsObject
> = (
  ...listenerArg: [
    ...EventArgsArray,
    TypedEvent<EventArgsArray & EventArgsObject>
  ]
) => void;
```

zemse:
Closes #335

### DIFF
--- a/packages/target-ethers-v5-test/types/commons.ts
+++ b/packages/target-ethers-v5-test/types/commons.ts
@@ -5,7 +5,7 @@
 import { EventFilter, Event } from "ethers";
 import { Result } from "@ethersproject/abi";
 
-export interface TypedEventFilter<EventArgsArray, EventArgsObject>
+export interface TypedEventFilter<_EventArgsArray, _EventArgsObject>
   extends EventFilter {}
 
 export interface TypedEvent<EventArgs extends Result> extends Event {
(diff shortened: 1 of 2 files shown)


krzkaczor:
@zemse I think you forgot to commit changed files. Running `yarn test:fix` locally and committing new files should do the trick. 

Maybe we can enable `noUnusedParameters` in our tsconfig so we can reproduce this problem in the future?

zemse:
> Maybe we can enable noUnusedParameters in our tsconfig so we can reproduce this problem in the future?

I tried setting in the global tsconfig, but it gives couple more errors. So lets do in another PR.

PR #340, opened 2021-03-08
docs: use @typechain/hardhat in hardhat.config.ts

roderik:
The old way only works when upgrading and not clearing out your npm_modules (as the old package is still installed)


krzkaczor:
🤦 thanks!

PR #345, opened 2021-03-12
fix: use CallOverrides, which is more complete than PayableOverrides

liamzebedee:
## Temporary fix
ethers.js allows users to pass the `from` field in overrides, which is useful for contracts like `WETH9.deposit({ from: accounts[3], value: '1' })`.

Ideally, this block should be reworked, as **all** ethers.js function calls are able to accept the `from` parameter as an override (as [shown here](https://sourcegraph.com/github.com/ethers-io/ethers.js@6c43e20e7a68f3f5a141c74527ec63d9fe8458be/-/blob/packages/contracts/src.ts/index.ts#L153:8)).

```
    !options.isStaticCall && !isConstant(fn) && !isConstantFn(fn)
      ? `overrides?: ${isPayable(fn) ? 'PayableOverrides' : 'Overrides'}`
      : 'overrides?: CallOverrides'
    'overrides?: CallOverrides'
```

### DIFF
--- a/packages/target-ethers-v5/src/codegen/functions.ts
+++ b/packages/target-ethers-v5/src/codegen/functions.ts
@@ -28,7 +28,7 @@ function generateFunction(options: GenerateFunctionOptions, fn: FunctionDeclarat
   ${generateFunctionDocumentation(fn.documentation)}
   ${overloadedName ?? fn.name}(${generateInputTypes(fn.inputs)}${
     !options.isStaticCall && !isConstant(fn) && !isConstantFn(fn)
-      ? `overrides?: ${isPayable(fn) ? 'PayableOverrides' : 'Overrides'}`
+      ? `overrides?: ${isPayable(fn) ? 'CallOverrides' : 'Overrides'}`
       : 'overrides?: CallOverrides'
   }): ${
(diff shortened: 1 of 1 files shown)


quezak:
Using `CallOverrides` is also invalid because calling a payable function shouldn't accept a `blockTag` override -- if we want to fix anything on our side we should create a separate interface that's correct.

If indeed all ethers calls accept a `from` in overrides param, the correct solution would be to post an issue in Ethers to add `from` to their `PayableOverrides` and/or `Overrides` interfaces.

krzkaczor:
This is needed indeed. `connect` doesn't work in the same way in some cases. I will try to figure out a fix in a sec.

krzkaczor:
For now a quick workaround would be to generate type with `& { from: string | Promise<string> }`.

I will create a separate issue to fix this (and few other things) in ethers.

krzkaczor:
Superseded by https://github.com/ethereum-ts/TypeChain/pull/350. Thanks for the original PR

liamzebedee:
Ah yeah, this is a better temporary solution @krzkaczor. Cheers!

Issue #330, opened 2021-02-26
generated typescript typing broken: must be unknown first


robert-e-davidson3:
When building the generated typescript, I get this error: 

```
TS2352: Conversion of type 'Contract' to type 'AccountLoader' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
```

Generated code:

```
/* Autogenerated file. Do not edit manually. */
/* tslint:disable */
/* eslint-disable */

import { Contract, Signer } from "ethers";
import { Provider } from "@ethersproject/providers";

import type { AccountLoader } from "../AccountLoader";

export class AccountLoader__factory {
  static connect(
    address: string,
    signerOrProvider: Signer | Provider
  ): AccountLoader {
    return new Contract(address, _abi, signerOrProvider) as AccountLoader;
  }
}

const _abi = <omitted for concision>
```

node version v12.19.0
ts-node version v9.0.0
yarn version 1.22.5

robert-e-davidson3:
nevermind, typescript version incompatibility

robert-e-davidson3:
This is still a problem with typescript version 4.2.2.

quezak:
@robert-e-davidson3 most likely the code is correct but there were some incompatible changes in ethers. Can you update all ethers packages and try again? See also a similar issue in #328

robert-e-davidson3:
Upgrading ethers 5.0.15 -> 5.0.31 fixed the issue!

quezak:
@krzkaczor we should bump the minimum ethers version required so this doesn't happen. IIRC it's 5.0.17 or 19

krzkaczor:
It's broken again on `5.0.32`.

I've submitted a PR that should improve this situation: https://github.com/ethereum-ts/TypeChain/issues/330

robert-e-davidson3:
You linked this issue. PR: https://github.com/ethereum-ts/TypeChain/pull/347

PR #347, opened 2021-03-14
Solve minor incompatibilities between ethers versions

krzkaczor:
CC @quezak @zemse 

Fixes: https://github.com/ethereum-ts/TypeChain/issues/330

### DIFF
--- a/.changeset/quiet-games-warn.md
+++ b/.changeset/quiet-games-warn.md
@@ -1,0 +1,7 @@
+---
+'@typechain/ethers-v5': patch
+---
+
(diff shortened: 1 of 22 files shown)


quezak (approved):
Importing from `ethers` namespace only is a good idea, but I'm afraid the `as any` casts will bite us at some point, when ethers actually change something. Werent' the inconsistencies caused only by using the explicit `@ethersproject/...` imports?

krzkaczor:
Good point - avoiding any casts should be fine.

PR #350, opened 2021-03-15
Allowed passing `from` as part of the overrides to contract call

krzkaczor:


packages/target-ethers-v5/src/codegen/index.ts
```diff
@@ -131,7 +131,11 @@ export function codegenContractTypings(contract: Contract) {
 export function codegenContractFactory(contract: Contract, abi: any, bytecode?: BytecodeWithLinkReferences): string {
   const constructorArgs =
     (contract.constructor && contract.constructor[0] ? generateInputTypes(contract.constructor[0].inputs) : '') +
-    `overrides?: ${contract.constructor[0]?.stateMutability === 'payable' ? 'PayableOverrides' : 'Overrides'}`
+    `overrides?: ${
+      contract.constructor[0]?.stateMutability === 'payable'
+        ? 'PayableOverrides & { from?: string | Promise<string> }'
+        : 'Overrides & { from?: string | Promise<string> }'
+    }`
```
quezak, line 138: let's add a 
```typescript
// TODO remove workaround for <ethers issue link>
```

quezak (approved):

PR #352, opened 2021-03-16
add example-hardhat-truffle

ducquangkstn:


krzkaczor:
Thanks for this example! Up until very recently, I wasn't even aware that people use hardhat with truffle 😅 

I added some tweaks in: https://github.com/ethereum-ts/TypeChain/pull/358

Issue #353, opened 2021-03-16
[ethers-v5] Error generating code for events with tuples


miltonjonat:
When an event definition includes a tuple among its arguments, the generated code in the corresponding `.d.ts` file is invalid for the ethers-v5 target. 

I have reproduced this problem using the latest code from the master branch (matches tag `@typechain/ethers-v5@6.0.3`). I have simply added an event with a tuple to the `examples/ethers-v5/abi/dai.json` file, as shown in this [fork's commit](https://github.com/cartesi/TypeChain/commit/b81de5ebf4af67a06e074845b0c08bd7ce4dc382).

Namely, I included an event definition as such in `examples/ethers-v5/abi/dai.json`:

```javascript
  {
    "anonymous": false,
    "inputs": [
      { "indexed": true, "internalType": "address", "name": "src", "type": "address" },
      {
        "indexed": false,
        "name": "data",
        "type": "tuple",
        "components": [
          { "name": "index", "type": "uint256" },
          { "name": "name", "type": "string" }
        ]
      }
    ],
    "name": "Tuple",
    "type": "event"
  }
```

Then the generated code in the `examples/ethers-v5/types/ethers-contracts/Dai.d.ts` is:

```javascript
    Tuple(
      src: string | null,
      data: null
    ): TypedEventFilter<
      [string, [BigNumber, string], { index: BigNumber; name: string }] & {
        src: string;
        data: [BigNumber, string] & { index: BigNumber; name: string };
      }
    >;
```

Which leads to the following error when running `yarn typecheck` within `examples/ethers-v5`
```
$ yarn typecheck
yarn run v1.22.5
$ tsc --noEmit
types/ethers-contracts/Dai.d.ts:837:8 - error TS2314: Generic type 'TypedEventFilter<EventArgsArray, EventArgsObject>' requires 2 type argument(s).

837     ): TypedEventFilter<
           ~~~~~~~~~~~~~~~~~
838       [string, [BigNumber, string], { index: BigNumber; name: string }] & {
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
... 
841       }
    ~~~~~~~
842     >;
    ~~~~~


Found 1 error.

error Command failed with exit code 2.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

[...]

PR #354, opened 2021-03-16
Fix code generation for events with tuples

krzkaczor:
Closes: https://github.com/ethereum-ts/TypeChain/issues/353

### DIFF
--- a/packages/target-ethers-v4-test/types/Events.d.ts
+++ b/packages/target-ethers-v4-test/types/Events.d.ts
@@ -24,6 +24,8 @@ interface EventsInterface extends Interface {
     emit_event3_overloaded: TypedFunctionDescription<{
       encode([]: []): string;
     }>;
+
+    emit_event4: TypedFunctionDescription<{ encode([]: []): string }>;
   };
 
   events: {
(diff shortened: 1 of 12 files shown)


packages/target-ethers-v4-test/types/Events.d.ts
```diff
@@ -41,6 +43,8 @@ interface EventsInterface extends Interface {
       encodeTopics([value1, value2]: [boolean | null, null]): string[];
     }>;
 
+    Event4: TypedEventDescription<{ encodeTopics([data]: [null]): string[] }>;
```
quezak, line 46: `data` of type `null`? That doesn't sound right 😄

quezak: or is this case just unsupported in ethers v4?

quezak (approved):
code looks ok at first glance 👌🏻

packages/target-ethers-v4-test/types/Events.d.ts
krzkaczor: Yeah I think it might not work with `ethers-v4`. Dunno - we don't care too much about that target.

Issue #379, opened 2021-05-06
Event types


janklimo:
Given this ABI:

```json
  {
    "name": "CheckpointToken",
    "inputs": [
      { "type": "uint256", "name": "time", "indexed": false },
      { "type": "uint256", "name": "tokens", "indexed": false }
    ],
    "anonymous": false,
    "type": "event"
  }
```

the generated type is

```ts
    CheckpointToken(
      time: null,
      tokens: null
    ): TypedEventFilter<
      [BigNumber, BigNumber],
      { time: BigNumber; tokens: BigNumber }
    >;
```

which only allows the following usage:

```ts
contract.filters.CheckpointToken(null, null)
```

As mentioned in [this comment](https://github.com/ethereum-ts/TypeChain/pull/321#issuecomment-785638289) `ethers.js` is fine accepting no arguments. I'm using the version `6.0.5` of the package and this isn't currently possible.

zemse:
@janklimo Thanks for the reminder, I'll put this on my to-do list for this weekend.

PR #380, opened 2021-05-09
[ethers-v5] Make filter parameters optional

zemse:
#379

Before:

```ts
Transfer(
  from: string | null,
  to: string | null,
  value: null
): TypedEventFilter<
  [string, string, BigNumber],
  { from: string; to: string; amount: BigNumber; }
>;

// for all erc20 transfers
const filter = contract.filters.Transfer(null, null, null)
```

After

```ts
Transfer(
  from?: string | null,
  to?: string | null,
  value?: null
): TypedEventFilter<
  [string, string, BigNumber],
  { from: string; to: string; amount: BigNumber; }
>;

// for all erc20 transfers
const filter = contract.filters.Transfer()

// for all erc20 transfers from a given address 0x1234...
const filter = contract.filters.Transfer('0x1234...')

// for all erc20 transfers to a given address 0x1234...
const filter = contract.filters.Transfer(null, '0x1234...')
```

### DIFF
--- a/packages/target-ethers-v5-test/test/Events.test.ts
+++ b/packages/target-ethers-v5-test/test/Events.test.ts
@@ -32,6 +32,19 @@ describe('Events', () => {
     })
   })
 
+  it('queryFilter without params', async () => {
+    await contract.emit_event1()
+
+    const filter = contract.filters.Event1()
+    const results = await contract.queryFilter(filter)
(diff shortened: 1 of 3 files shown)


krzkaczor:
Thanks, I will add a changeset in a separate commit. I am planning to release everything today.

PR #388, opened 2021-05-24
update README with ethers-v5

zhangyuannie:
It is better to add it since user might think `ethers-v5` is not supported.


krzkaczor:
Makes sense, thanks!

Issue #367, opened 2021-04-07
Hardhat with external dependency


ducquangkstn:
Hi I am using typeChain with hardhat.
I try to import dependency (json file).

Is there any way to include json path (in node_modules folder) when generating typeChain files

krzkaczor:
I think currently there is no way to achieve it. Thanks for raising this. I will make sure to add it in: https://github.com/ethereum-ts/TypeChain/pull/363

krzkaczor:
btw. as a quick workaround you can always run TypeChain again as a separate build step (and for example set `outDir` to something else). Ugly but it should work.

ducquangkstn:
Yeah, right now, I am doing that way.
The thing is it will create 2 typechain folders (1 for my repo and 1 for the library), so it is kind of ugly solution

krzkaczor:
Sorry, I decided to drop it from the release. Feel free to pick it up.

moodysalem:
people who use the Uniswap V3 npm artifacts probably want this

omarish:
> people who use the Uniswap V3 npm artifacts probably want this

+1. I ended up doing this as a workaround:

https://github.com/omarish/uniswap-v3-staker/blob/55c0b03f68eef947b41205ce6623c0f3f13292ad/hardhat.config.ts#L9-L32

ducquangkstn:
> people who use the Uniswap V3 npm artifacts probably want this

Actually, I work at Kyber. :))))
But we built a lot of bridge reserve and tried to publish our artifacts as well

krzkaczor:
The good news is that @ducquangkstn 's fix looks good. Just needs a little bit of tweaking so this should be done soon.

PR #394, opened 2021-06-02
Hardhat allows external artifacts

ducquangkstn:
Resolve https://github.com/ethereum-ts/TypeChain/issues/367


packages/hardhat/src/index.ts
```diff
@@ -43,15 +44,19 @@ subtask(TASK_COMPILE_SOLIDITY_COMPILE_JOBS, 'Compiles the entire project, buildi
...
-    const needsFullRebuild = taskArgsStore.fullRebuild || typechainCfg.target !== 'ethers-v5'
+    const needsFullRebuild =
+      taskArgsStore.fullRebuild || typechainCfg.target !== 'ethers-v5' || typechainCfg.externalArtifacts
     console.log(
       `Generating typings for: ${artifactPaths.length} artifacts in dir: ${typechainCfg.outDir} for target: ${typechainCfg.target}`,
     )
     const cwd = process.cwd()
-    const { runTypeChain, glob } = await import('typechain')
-    const allFiles = glob(cwd, [`${config.paths.artifacts}/!(build-info)/**/+([a-zA-Z0-9_]).json`])
+    const artifactsPaths = [`${config.paths.artifacts}/!(build-info)/**/+([a-zA-Z0-9_]).json`]
```
krzkaczor, line 55: I think the better implementation would be to first run glob ignoring node_modules and later again to get external artifacts. Finally, merge both results. This way we don't modify old behaviour at all.

packages/hardhat/src/index.ts
```diff
@@ -43,15 +44,19 @@ subtask(TASK_COMPILE_SOLIDITY_COMPILE_JOBS, 'Compiles the entire project, buildi
 
     // RUN TYPECHAIN TASK
     const typechainCfg = config.typechain
-    // incremental generation is only supported in 'ethers-v5'
+    // incremental generation is only supported in 'ethers-v5' and external artifacts
     // @todo: probably targets should specify somehow if then support incremental generation this won't work with custom targets
-    const needsFullRebuild = taskArgsStore.fullRebuild || typechainCfg.target !== 'ethers-v5'
+    const needsFullRebuild =
```
krzkaczor, line 49: Can we somehow support incremental generation with `externalArtifacts`? If I get it correctly we have no way of checking if we already generated types for externalArtifacts and this is the reason for this change. But maybe we can always generate typings only for `externalArtifacts` and avoid regenerating all types for internal files? Incremental generation is often a lifesaver.

krzkaczor (changes requested):
Thanks for working on it! Some comments.

ducquangkstn:
@krzkaczor can you review this PR again?

krzkaczor (approved):

krzkaczor:
Thanks! I will add changesets and try to release it soon.

krzkaczor:
Released as `@typechain/hardhat@2.1.0`

Issue #403, opened 2021-06-16
Feature Request: Include source file(s) in the preamble for output .d.ts files


NiloCK:
This project goes a long way in improving the readability / **discoverability** of TS wrapped solidity code. A small improvement could be to include references to the solidity implementations in the generated .d.ts files.

A draft implementation looks like this:

`packages/typechain/src/outputTransformers/sources.ts`: (new file)

``` ts
import { OutputTransformer } from '.'

export const sourceFilesOutputTransformer: OutputTransformer = (output, services, config) => {
  const filesList = config.filesToProcess.map( f => ` * ${f}` )

  return [
    '/** implementation of these types can be found in source files: ',
    ...filesList,
    ' */',
    output
  ].join('\n')
}
```

`packages/typechain/src/outputTransformers/index.ts`: (add the new transformer)

``` ts
import { Config, Services } from '../../typechain/types'
import { addPreambleOutputTransformer } from './preamble'
import { prettierOutputTransformer } from './prettier'
import { sourceFilesOutputTransformer } from './sources'

export type OutputTransformer = (output: string, services: Services, cfg: Config) => string

export const outputTransformers = [
  sourceFilesOutputTransformer,
  addPreambleOutputTransformer,
  prettierOutputTransformer
]
```

Problems here are:
- this seems to be printing *every* solidity source file to every .d.ts file. I've misunderstood `.filesToProcess`?
- the listed files point to the contract abi json rather than the original .sol sources
- the listed files contain the absolute system path, which will be different from user to user, so introduces a problem for the existing testing format

Thoughts?

Thanks for the great project!

PR #404, opened 2021-06-16
Include Source Files in the preamble for output .d.ts files

NiloCK:
This is in reference to issue #403. It adds information to the preamble of generated files in order to point developers from the generated d.ts files to the correct implementing solidity code.

The actual code changes are in the first three commits and are quite small - the large diff comes from updating the generated outputs from the test suite.

This is a first-pass with potential warts, but I think it represents an improvement to the project as is. Someone who knows the project could feel free to inform me why *some* reported source files are `.json` and others are `.abi`.

Better still would be to move the determination of `Config.sourceFile` up the chain somewhat, so that it referenced the actual solidity file itself rather than the built artifacts.


NiloCK:
(It occurs to me that maybe TypeChain doesn't know anything about the originating solidity implementation files, and instead begins its type generation quest by searching for the build artifacts themselves). In which case, the existing pointers toward the built artifacts will have to do!

krzkaczor:
Hey @NiloCK.

> (It occurs to me that maybe TypeChain doesn't know anything about the originating solidity implementation files, and instead begins its type generation quest by searching for the build artifacts themselves). 

Yeah that's true. TypeChain takes json/abi files as input not solidity files. 

I am not 100% sure if this is needed... I personally don't see too much value for such feature. 

If you're a TC user and want to see this change please speak up. For now, I will keep this opened.

krzkaczor:
Sorry, I am gonna close it for now. I don't feel like this is a much needed feature.

PR #407, opened 2021-06-22
Add Kyber to logo show-case in README

ayobuenavista:
Kyber is a huge user of TypeChain, and we have it in a lot of our projects. If it's possible, it would be great to add Kyber's logo as a show of support to TypeChain.


krzkaczor:
It's great to have you among TC users ;)

PR #414, opened 2021-06-24
Add missing eslint rules to find missing runtime deps

krzkaczor:


packages/hardhat/package.json
```diff
@@ -40,10 +40,12 @@
...
     "rimraf": "^3.0.2",
-    "typechain": "*"
+    "typechain": "*",
+    "@types/lodash": "^4.14.139"
   },
   "peerDependencies": {
     "hardhat": "^2.0.10",
-    "typechain": "^5.0.0"
+    "typechain": "^5.0.0",
+    "lodash": "^4.17.15"
```
freeatnet, line 49: Hey @krzkaczor, it looks like this new peer dependency isn't mentioned in the docs, should it be included in the `npm install --save-dev` command?

krzkaczor: You don't have to add it on your end -- `lodash` is a `dependency` of typechain.

Issue #420, opened 2021-07-01
[hardhat] Type extensions not working


Mrtenz:
It looks like the type extensions for `@typechain/hardhat` aren't working, resulting in a type error when trying to use `typechain` inside the Hardhat configuration.

Simply importing the plugin like
```ts
import '@typechain/hardhat';
```

works fine, but trying to configure TypeChain like this for example

```ts
const config: HardhatUserConfig = {
  typechain: {
    outDir: 'contracts',
  }
}
```

results in an error:

```
Object literal may only specify known properties, and 'typechain' does not exist in type 'HardhatUserConfig'.
```

The [`type-extensions.ts`](https://github.com/ethereum-ts/TypeChain/blob/master/packages/hardhat/src/type-extensions.ts) is not imported in [`index.ts`](https://github.com/ethereum-ts/TypeChain/blob/master/packages/hardhat/src/index.ts) which could be the cause, however I tried adding it locally and that did not solve the issue.

krzkaczor:
Probably it's some kind of details in tsconfig. For reference you can use this repo: https://github.com/ethereum-ts/TypeChain/tree/master/examples/hardhat

I think you are right that `type-extensions.ts`, should be explicitly required. If you know how other plugins do it, feel free to submit a PR. 

An obvious, dirty workaround would be not to type config as `HardhatUserConfig` for now ...

Mrtenz:
Adding `typechain` in the config in the example results in the same error, but it seems like that one can be fixed by importing `type-extensions.ts`. [This is what Hardhat itself does as well](https://github.com/nomiclabs/hardhat/blob/877c67815de03c436eb1ae9d11739f82791bb665/packages/hardhat-ethers/src/internal/index.ts#L12), so I'll submit a PR for this.

> An obvious, dirty workaround would be not to type config as HardhatUserConfig for now ...

Sure, it would of course be better to have proper type checking in the config.

krzkaczor:
It's strange because it generally works... I tested it in multiple projects and people use it for a few weeks now.

stevyhacker:
I'm having the same error message as above.

krzkaczor:
Released as: https://github.com/ethereum-ts/TypeChain/releases/tag/%40typechain%2Fhardhat%402.1.2

Let me know if it works

ageyev:
> Let me know if it works

No. Hardhat code should be fixed (see pull request to Hardhat repo above)

krzkaczor:
[...]

PR #424, opened 2021-07-02
Import type-extensions in Hardhat plugin index

Mrtenz:
This at least partially fixes the issue described in #420. The `HardhatPluginError` import was unused so I removed that, and added the `./type-extensions` import to make sure it's included in the resulting `index.d.ts`.


krzkaczor:
Thanks for this. I am gonna improve linter setup to catch unused imports. Once I am done I am gonna release this change.

PR #439, opened 2021-08-02
fix: remove double "await" in hardhat plugin tests

paulrberg:


krzkaczor:
Thanks! This would be normally caught by our liniting but `.run` returns promise of any 😬

PR #441, opened 2021-08-02
docs: add @paulrberg's solidity-template example

paulrberg:
One of the main attractions of my [solidity-template](https://github.com/paulrberg/solidity-template) is the off-the-shelf integration of TypeChain, and of this Hardhat plugin.


krzkaczor:
Thanks! This is a great addition indeed.

PR #451, opened 2021-08-12
fix: Add EIP1559 fee parameters

rafinskipg:
Types are not being generated with the new EIP1559 fees, I hope this fix is correct. 🤕

### DIFF
--- a/packages/target-truffle-v4/static/types.d.ts
+++ b/packages/target-truffle-v4/static/types.d.ts
@@ -40,6 +40,8 @@ declare namespace Truffle {
     from?: string
     gas?: BigNumber | number | string
     gasPrice?: BigNumber | number | string
+    maxPriorityFeePerGas?: string | number | BN
+    maxFeePerGas?: string | number | BN
     value?: BigNumber | string
   }
 
--- a/packages/target-truffle-v5-test/types/truffle-contracts/types.d.ts
+++ b/packages/target-truffle-v5-test/types/truffle-contracts/types.d.ts
@@ -62,6 +62,8 @@ declare namespace Truffle {
     from?: string;
     gas?: BN | number | string;
     gasPrice?: BN | number | string;
+    maxPriorityFeePerGas?: string | number | BN;
+    maxFeePerGas?: string | number | BN;
     value?: BN | string;
   }
 
--- a/packages/target-ethers-v4-test/types/index.d.ts
+++ b/packages/target-ethers-v4-test/types/index.d.ts
@@ -10,6 +10,8 @@ import {
 export interface TransactionOverrides {
   gasLimit?: BigNumberish | Promise<BigNumberish>;
   gasPrice?: BigNumberish | Promise<BigNumberish>;
+  maxPriorityFeePerGas?: BigNumberish | Promise<BigNumberish>;
+  maxFeePerGas?: BigNumberish | Promise<BigNumberish>;
   nonce?: BigNumberish | Promise<BigNumberish>;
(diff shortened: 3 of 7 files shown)


krzkaczor:
Hey, thanks for this PR. I assume that support for eip 1559 landed in ethers already. Probably we should try to reuse types exposed by ethers rather than redefining them... i don't remember what was the reasoning for this. Does eip1559 work in ethers-v4 too?

rafinskipg:
@krzkaczor  I didn't try ethers, i just realized that the new types were not being generated locally so i just added them. Types were added in web3-js 1.5.1 but i don't know about ethers

packages/target-truffle-v4/static/types.d.ts
```diff
@@ -40,6 +40,8 @@ declare namespace Truffle {
     from?: string
     gas?: BigNumber | number | string
     gasPrice?: BigNumber | number | string
+    maxPriorityFeePerGas?: string | number | BN
+    maxFeePerGas?: string | number | BN
```
anthanh, line 44:
```suggestion
    maxFeePerGas?: BigNumber | number | string
```

packages/target-truffle-v4/static/types.d.ts
```diff
@@ -40,6 +40,8 @@ declare namespace Truffle {
     from?: string
     gas?: BigNumber | number | string
     gasPrice?: BigNumber | number | string
+    maxPriorityFeePerGas?: string | number | BN
```
anthanh, line 43:
```suggestion
    maxPriorityFeePerGas?: BigNumber | number | string
```

packages/target-truffle-v5-test/types/truffle-contracts/types.d.ts
```diff
@@ -62,6 +62,8 @@ declare namespace Truffle {
     from?: string;
     gas?: BN | number | string;
     gasPrice?: BN | number | string;
+    maxPriorityFeePerGas?: string | number | BN;
+    maxFeePerGas?: string | number | BN;
```
anthanh, line 66:
```suggestion
    maxFeePerGas?: BN | number | string;
```
Same order like others.

packages/target-truffle-v5-test/types/truffle-contracts/types.d.ts
```diff
@@ -62,6 +62,8 @@ declare namespace Truffle {
     from?: string;
     gas?: BN | number | string;
     gasPrice?: BN | number | string;
+    maxPriorityFeePerGas?: string | number | BN;
```
anthanh, line 65:
```suggestion
    maxPriorityFeePerGas?: BN | number | string;
```
Same order like others

anthanh (commented):
Actually, looks like using ethers exported types, but checks fails at truffle-v4 because of a missing type.

packages/target-truffle-v5-test/types/truffle-contracts/types.d.ts
anthanh: wdyt @krzkaczor ?

krzkaczor:
Thanks for working on this. I will make sure to merge this during the weekend. It probably needs some runtime tests as well...

krzkaczor:
New PR: https://github.com/ethereum-ts/TypeChain/pull/456#pullrequestreview-736480312 (h/t @zemse)

Thanks for the initial implementation! In the WIP PR we try to rely on ethers exported types instead of redeclaring them. It should be merged soon.

Issue #440, opened 2021-08-02
Generate and export named event responses for ethers v5


Siegrift:
Currently TypeChain generates types similar to the following:

```ts
class ContractAbc extends Contract {
...
  filters: {
    TemplateCreated(
      templateId: BytesLike | null,
      airnodeId: null,
      endpointId: null,
      parameters: null
    ): TypedEventFilter<
      [string, string, string, string],
      {
        templateId: string;
        airnodeId: string;
        endpointId: string;
        parameters: string;
      }
    >;
    ...
```

The problem is that I can't use filters to query my event. Instead I want to pass `tx.hash`, and parse the logs myself. However at that point I would like to cast my event response to the Type generated by TypeChain.

For example, this code:
```ts
    contractAbc.provider.once(tx.hash, (tx) => {
      const parsedLog = airnodeRrp.interface.parseLog(tx.logs[0]);
      resolve(parsedLog.args.templateId); // This is untyped (args is basically any)
    })
```

Instead I would like to have a code like this:
```ts
    contractAbc.provider.once(tx.hash, (tx) => {
      const parsedLog = airnodeRrp.interface.parseLog(tx.logs[0]).args as TemplateCreatedEvent;
      resolve(parsedLog.templateId); // parsedLog should be typed
    })
```

Slightly related: https://github.com/ethereum-ts/TypeChain/issues/249

StErMi:
It would be cool to have the possibility to explicitly say in the typechain export moment which interface use for types so also those interfaces would be exported and be usable on the frontend side directly

krzkaczor:
Done in https://github.com/dethcrypto/TypeChain/releases/tag/%40typechain%2Fethers-v5%407.1.0

Duplicate of: https://github.com/dethcrypto/TypeChain/issues/398

StErMi:
damn cool!

PR #457, opened 2021-08-22
[ethers-v5] Generate and export typed events

zemse:
Usage

```ts
import type { Transfer_TypedEvent } from '../typechain/MyErc20Token';

// overloaded events
import type { Execute_bytes32_TypedEvent, Execute_address_bytes_TypedEvent } from '../typechain/MyMultisig';

// ...

const parsed = contract.interface.parseLog(log) as TransferEvent;
parsed.args.from // string
parsed.args.to // string
parsed.args.value // BigNumber
```

#440

### DIFF
--- a/.changeset/good-starfishes-appear.md
+++ b/.changeset/good-starfishes-appear.md
@@ -1,6 +1,5 @@
 ---
 '@typechain/ethers-v5': minor
-'@typechain/ethers-v5-test': minor
 ---
 
 Export typed events
--- a/.changeset/good-starfishes-appear.md
+++ b/.changeset/good-starfishes-appear.md
@@ -1,0 +1,6 @@
+---
+'@typechain/ethers-v5': minor
+'@typechain/ethers-v5-test': minor
+---
+
+Export typed events
--- a/packages/target-ethers-v5-test/types/Events.d.ts
+++ b/packages/target-ethers-v5-test/types/Events.d.ts
@@ -221,9 +221,21 @@ export class Events extends BaseContract {
   };
 
   filters: {
+    "AnonEvent1(uint256)"(
+      value1?: BigNumberish | null
+    ): TypedEventFilter<[BigNumber], { value1: BigNumber }>;
+
     AnonEvent1(
       value1?: BigNumberish | null
     ): TypedEventFilter<[BigNumber], { value1: BigNumber }>;
+
+    "Event1(uint256,uint256)"(
+      value1?: BigNumberish | null,
+      value2?: null
+    ): TypedEventFilter<
+      [BigNumber, BigNumber],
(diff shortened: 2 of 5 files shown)


.changeset/good-starfishes-appear.md
```diff
@@ -0,0 +1,6 @@
+---
+'@typechain/ethers-v5': minor
+'@typechain/ethers-v5-test': minor
```
krzkaczor, line 3: We don't version `-test` packages as they are not published. You can simply remove this line.

krzkaczor (commented):
1. Does this implementation work with overloaded events (events with the same name but different signatures)? I don't think so as there would be a name conflict. I think we need to dump the whole signature into the type name if we detect multiple events with the same name (I think that there is already a function that does this).

2. I think it makes sense to use these pregenerated exported types in the event filters code here: https://github.com/ethereum-ts/TypeChain/pull/457/files#diff-a783a7ebd19f6659f4d15eafcb5014e2881776e245845089383b47e5b4758f99R231

packages/target-ethers-v5-test/types/Events.d.ts
```diff
@@ -93,6 +93,32 @@ interface EventsInterface extends ethers.utils.Interface {
   getEvent(nameOrSignatureOrTopic: "NoArgsEvent"): EventFragment;
 }
 
+export type AnonEvent1_TypedEvent = TypedEvent<
+  [BigNumber] & { value1: BigNumber }
+>;
+
+export type Event1_TypedEvent = TypedEvent<
```
krzkaczor, line 100: Can we simply use `AnonEvent1Event` instead of `AnonEvent1_TypedEvent` in a happy case? Where there are overloads let's use `Event3_bool_uint256_Event` instead of `Event3_bool_uint256_TypedEvent`. So basically just drop `Typed`.

packages/target-ethers-v5-test/types/Events.d.ts
```diff
@@ -201,10 +227,22 @@ export class Events extends BaseContract {
   };
 
   filters: {
+    "AnonEvent1(uint256)"(
+      value1?: BigNumberish | null
+    ): TypedEventFilter<[BigNumber], { value1: BigNumber }>;
```
krzkaczor, line 232: Here it would be perfect to reuse the already generated type ie: 
`TypedEventFilter<AnonEvent1Event>;` but I think it's not that simple so we can ship initial implementation without this and later improve.

zemse: I agree we should, that may involve some refactor to the TypedEventFilter generics. Maybe we can clean up the code in another PR.

PR #479, opened 2021-09-24
[ethers-v5] fix for getContractAt incorrect type

zemse:
The PR https://github.com/dethcrypto/TypeChain/pull/473 introduced a bug that the getContractAt returns a factory type. But it should return a contract type instead. This PR aims to fix this.

### DIFF
--- a/.changeset/unlucky-ties-beg.md
+++ b/.changeset/unlucky-ties-beg.md
@@ -1,0 +1,5 @@
+---
+'@typechain/ethers-v5': patch
+---
+
+Fix typing for `getContractAt` when using Hardhat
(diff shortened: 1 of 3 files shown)


cinnabarhorse:
Our pipeline was directly affected by this bug -- would appreciate a quick merge!

krzkaczor:
@cinnabarhorse releasing it now but for a quick fix just downgrade to the previous version. Later today we will improve our testing pipeline so this never happens again. Sorry.

krzkaczor (approved):

krzkaczor:
Released as `@typechain/ethers-v5@7.1.2`

PR #486, opened 2021-10-01
Basic works-for-me hack to use .ts interfaces rather than .d.ts

silasdavis:
declarations so that compiles includes type declarations in build
output of typechain-based projects.

https://github.com/dethcrypto/TypeChain/issues/430

Signed-off-by: Silas Davis <silas@monax.io>


krzkaczor:
Thanks for this @silasdavis . Track progress on this on: https://github.com/dethcrypto/TypeChain/pull/504

PR #488, opened 2021-10-01
[ethers-v5] Function fragments tuples

zemse:


packages/target-ethers-v5/src/codegen/functions.ts
```diff
@@ -82,7 +82,7 @@ function generateFunctionDocumentation(doc?: FunctionDocumentation): string {
 }
 
 export function generateInterfaceFunctionDescription(fn: FunctionDeclaration): string {
-  return `'${generateFunctionSignature(fn)}': FunctionFragment;`
+  return `'${getSignatureForFn(fn)}': FunctionFragment;`
 }
 
 export function generateFunctionSignature(fn: FunctionDeclaration): string {
```
krzkaczor, line 88: Please remove `generateFunctionSignature` as well as it's not used anymore.

krzkaczor (changes requested):

PR #492, opened 2021-10-06
Use local file paths in examples dependencies

hasparus:
### What's changed?

- Examples now depend on TypeChain packages using `file:../...` relative paths.
- I rewrote `check-examples.sh` to TypeScript, because it's a superior scripting language 😜 and I didn't know what's wrong when the bash script failed.

### Further work

"Why did the bash script fail?" you might ask.
In **truffle-v4** example, depending on **@typechain/truffle-v4** from the repository instead of `^2.0.3` failed with _"Couldn't find truffle-v4 error."_ ([screenshot](https://user-images.githubusercontent.com/15332326/136196173-176a3452-e92a-4947-84c3-862e6115579d.png)). I might take a look at it later, but it's out of scope of this PR.


scripts/check-examples.ts
```diff
@@ -0,0 +1,56 @@
+// @ts-check
```
krzkaczor, line 1: do we need this?

krzkaczor (approved):

scripts/check-examples.ts
```diff
@@ -0,0 +1,56 @@
...
+  }
+}
+
+if (failures.length > 0) {
+  console.error(
+    '\n',
+    bold(failures.length) + ` example${failures.length > 1 ? 's' : ''} failed:`,
+    red(failures.join(', ')),
+    '\n',
+  )
```
hasparus, line 43: I forgot to fail the process.

```suggestion
  )
  process.exit(1)
```

Issue #474, opened 2021-09-23
Improve testing pipeline


krzkaczor:
* Introduce popular contract's ABIs (uniswap, MCD?) as part of the pipeline
* Add new tests using newer solidity version (structs, errors?)

krzkaczor:
* always test examples with the local version of packages (typechain and targets)

PR #501, opened 2021-10-08
Test contracts compiled with many Solidity versions and handle reserved words

hasparus:
I decided https://github.com/dethcrypto/TypeChain/issues/497 is a good reason to test with multiple Soliditiy versions, so this PR aims to solve both the crash on reserved words and address #474.

### DIFF
--- a/.changeset/itchy-gifts-deliver.md
+++ b/.changeset/itchy-gifts-deliver.md
@@ -1,0 +1,5 @@
+---
+'typechain': minor
+---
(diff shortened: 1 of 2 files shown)


krzkaczor:
It's missing changesets. You can try a minor change in TC core and we see if changeset properly changes peer dependency on targets (now ethers-v5 will require at least `5.2`) I feel we will need to do a major release for that.

PR #504, opened 2021-10-11
Generate types to `.ts` instead of `.d.ts` files.

hasparus:
### What's changed

- Fixed warnings from Solidity files (tests printed scary yellow text and I got scared)
- Changed emitted files to `.ts` for Ethers v5 target.

### What needs to be discussed

I left `hardhat.d.ts` as `.d.ts` because after renaming to `.ts` extending **HardhatEthersHelpersBase** in hardhat fixture project clashes with the definition pf**HardhatEthersHelpers** of _@nomiclabs/hardhat-ethers._

```
Interface 'import("d:/workspace/TypeChain/node_modules/hardhat/types/runtime").HardhatEthersHelpers' incorrectly extends interface 'import("d:/workspace/TypeChain/node_modules/@nomiclabs/hardhat-ethers/types/index").HardhatEthersHelpers'.

 Types of property 'getContractFactory' are incompatible.
    Type '{ (name: "EdgeCases", signerOrOptions?: Signer | FactoryOptions | undefined): Promise<EdgeCases__factory>; (name: "SafeMath", signerOrOptions?: Signer | ... 1 more ... | undefined): Promise<...>; (name: "TestContract", signerOrOptions?: Signer | ... 1 more ... | undefined): Promise<...>; (name: "TestContract1", sign...' is not assignable to type '{ (name: string, signerOrOptions?: FactoryOptions | Signer | undefined): Promise<ContractFactory>; (abi: any[], bytecode: BytesLike, signer?: Signer | undefined): Promise<...>; }'.
      Types of parameters 'name' and 'name' are incompatible.
        Type 'string' is not assignable to type '"EdgeCases"'.ts(2430)
```

Thinking about it more, I feel leaving that modules augmentation in a `.d.ts` file might make sense — this is basically writing to global namespace, so if somebody publishes this generated file to somebody elses node_modules, we can easily end up with a clash on `getContractFactory` types.

### To-Do

- [x] Add a test to assert that `declare module "hardhat/runtime/types"` works.
- [x] Emit `.ts` in web3js target.

---

**Disclaimer:** This might have problems not covered by our testing pipeline, so I'd probably prerelease and try using it in real world projects.

### DIFF
--- a/examples/hardhat/package.json
+++ b/examples/hardhat/package.json
@@ -26,7 +26,6 @@
     "hardhat": "^2.0.10",
     "ts-node": "^9.1.1",
     "typechain": "file:../../packages/typechain",
-    "@typechain/test-utils": "file:../../packages/test-utils",
     "typescript": "^4.1.5"
   },
   "resolutions": {
--- a/examples/hardhat/test/counter.ts
+++ b/examples/hardhat/test/counter.ts
@@ -3,8 +3,7 @@ import chaiAsPromised from 'chai-as-promised'
 import { solidity } from 'ethereum-waffle'
 import { ethers } from 'hardhat'
 
-import type { Counter, Counter__factory } from '../typechain'
-import type { AssertTrue, IsExact } from '@typechain/test-utils'
+import type { Counter } from '../typechain'
 
 chai.use(solidity)
 chai.use(chaiAsPromised)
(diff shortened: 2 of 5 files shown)


paulrberg:
This is nice! This will be shipped with v6, won't it? What's the ETA for that?

krzkaczor:
@paulrberg next week 🔥

Jack-Works:
why? this is inconvenient in monorepo.

PR #514, opened 2021-10-12
Add KingOfTheHill contract with struct, error and event

hasparus:
I've added a new contract that's quite short but it has a little bit of everything. We can use it to test TypeChain's behaviour on Solidity `struct`s and `error`s. 

Part of https://github.com/dethcrypto/TypeChain/issues/474 I guess? Treat this PR as a point to discuss instead of a change I wanna make to the code.

**Idea:** We already emit some runtime for factories. Should we use the information about errors from the ABI to emit type guards for errors?

The following ABI

```json
  {
    "inputs": [
      { "internalType": "uint256", "name": "value", "type": "uint256" },
      {
        "internalType": "uint256",
        "name": "highestBidValue",
        "type": "uint256"
      }
    ],
    "name": "BidNotHighEnough",
    "type": "error"
  }
```

could become something like

```ts
interface BidNotHighEnough extends Error { /* ... */ }
function isBidNotHighEnough(error: unknown): error is BidNotHighEnough { /* ... */ }

try {
  await contract.bid({ value: 1 })
} catch (err) {
  if (isBidNotHighEnough(err)) {
    console.log(err.highestBidValue)
  }
}
```

(As a side note, we'd need to fiddle with "ethers" dependency a bit, because we currently encounter https://github.com/ethers-io/ethers.js/issues/1493)


krzkaczor:
@hasparus yes go ahead.

PR #517, opened 2021-10-12
Add notes about the need to build monorepo before running examples

krzkaczor:
Related: https://github.com/dethcrypto/TypeChain/issues/510


hasparus:
nit: I'd personally put this below _"## Running"_, but it's fine.

PR #518, opened 2021-10-13
Use previously defined event types in Ethers 5 target

hasparus:
Depends on https://github.com/dethcrypto/TypeChain/pull/504.

Changes how events related code is emitted in Ethers 5 target.

### DIFF
--- a/examples/hardhat/package.json
+++ b/examples/hardhat/package.json
@@ -26,7 +26,6 @@
     "hardhat": "^2.0.10",
     "ts-node": "^9.1.1",
     "typechain": "file:../../packages/typechain",
-    "@typechain/test-utils": "file:../../packages/test-utils",
     "typescript": "^4.1.5"
   },
   "resolutions": {


paulrberg:
Hey guys! I am in the process of upgrading my [solidity-template](https://github.com/paulrberg/solidity-template) to the latest version of TypeChain, which includes the changed applied via this PR.

I don't think I follow what it means for "named events to be used in Contract methods". I am also a bit lazy to go through the 37+ diffed files :D. Any chance you could share a basic example for what this PR means, in practice?

krzkaczor:
Sure! It's all about reducing repetition in generated code.

For some time now TypeChain exports types for events like: 
```ts
export type CountedToEvent = TypedEvent<[BigNumber], { number: BigNumber }>;
```

Despite having these types already defined, while defining filters we provided all types again instead of simply using them. So this:

```ts
filters: {
    "CountedTo(uint256)"(
      number?: null
    ): TypedEventFilter<[BigNumber], { number: BigNumber }>;
}
```

becomes now this:

```ts
filters: {
    "CountedTo(uint256)"(number?: null): TypedEventFilter<CountedToEvent>;
};
```

paulrberg:
Gotcha, thanks Kris!

PR #519, opened 2021-10-13
Use `status --verbose` instead of diff `--no-pager` in `no-git-changes.sh`

hasparus:
### Before

image

It seems that `git --no-pager diff` doesn't display on CI when new files are added.

### After

`git status --verbose --verbose` should display both new files and changes.

image


krzkaczor:
Just to make sure: double `--verbose` is not a type, right? :)

PR #520, opened 2021-10-15
Change default outDir of @typechain/hardhat to 'typechain-types'

hasparus:
Suprisingly, we using `.typechain` as outDir affects how user's tsconfig includes (globs seem to be required if directory starts with `"."` :>) so we'll go with `typechain-types`


hasparus:
Hey @krzkaczor, I think we don't wanna commit lockfiles in examples dir. It seem that running `yarn install` with `--frozen-lockfile` can get us an old version of local libraries (despite of `file:../` installation path).

image

paulrberg:
For posterity: see related discussion in https://github.com/dethcrypto/TypeChain/issues/423.

PR #543, opened 2021-11-09
Exclude unused imports from being added to the type file

freeatnet:
This PR attempts to tighten up the import lists and avoid importing things that won't be used in the type file. Fixes the issue mentioned here: https://github.com/dethcrypto/TypeChain/issues/430#issuecomment-964314162

Not super happy with the code, but it works. Let me know if there's anything that I need to change.


freeatnet:
Looks like I missed types used by `generateStruct`.

hasparus:
Nice one! Thanks for the PR, @freeatnet 🙏

PR #547, opened 2021-11-15
Collect unused imports in `ethers-v5`

hasparus:
### What's changed?

- Added `createImportsForUsedIdentifiers` syntax helper to `typechain`. It runs regexes on the entire source file, so... well, it's not my best work. If someone complains that TypeChain is slow we can move towards "ordering" imports when using identifiers. This should be fine for now, though.
- Used it in `ethers-v5` target.
- Enabled `exactOptionalPropertyTypes` TypeScript compiler flag. I tried to enable `noImplicitOverride`, but it's not so trivial — we'll do it separately.


krzkaczor (approved):
I agree that "ordering" imports is probably the way to go but this mechanism should be good enough. Keep in mind that factory codegen has also similar mechanism so it would be good to normalize it in the future.

PR #551, opened 2021-11-18
Fix constant size struct arrays in Ethers v5 target

### LINKED ISSUE #549: Bug: SyntaxError while generating typechain for projects
Hi there.

So I have been experimenting with typechain for the past few days, and discovered that the current version of typechains cannot properly compile my company's project, with the following error

image

However, if I revert all typechain-related packages to their latest versions before the added support for Solidity structs, everything will work normally. More specifically, the versions I reverted to were:
```
@typechain/hardhat@2.3.1
typechain@5.2.0
@typechain/ethers-v5@7.2.0
```

Here is how to recreate the error: 
- Clone [this repo](https://github.com/pendle-finance/pendle-core/tree/demo-issue-typechain)
- Run `yarn install`, then `yarn compile`

It would be super helpful if this error can be fixed! Thanks a lot!

hasparus:
Closes #549.

### What's changed?

- I fixed our handling of constant size struct arrays (e.g. `(PendingTransfer[3]` memory transfers)`), so they don't lead to incorrect TS emit anymore.
- Bumped TS version to 4.5. I hoped it can fix a problem I had locally, but it wasn't actually needed. Regardless, it's stable now, so why not keep it.
- Started using solc ^8 in hardhat tests — I wanted to test struct arrays in constructor, but in solc ^6 it fails with the error about AbiCoder version being too low :>

### What was wrong?

Example: In ABI input like the following...

```json
        {
          "components": [
            {
              "internalType": "uint256",
              "name": "x",
              "type": "uint256"
            },
            {
              "internalType": "uint256",
              "name": "y",
              "type": "uint256"
            }
          ],
          "internalType": "struct Vector2[2]",
          "name": "segment",
          "type": "tuple[2]"
        }
```

We treated `Vector2[2]` as the struct name.

### What can be improved here?

I can now see that it may have been possible to modify `parseEvmType` instead and limit the change to one function. I can give it a shot tomorrow.


krzkaczor (approved):
Missing changeset

krzkaczor (approved):

PR #557, opened 2021-11-24
Stop emitting overly long tuple types

hasparus:
Helps with OOM crashes and closes https://github.com/dethcrypto/TypeChain/issues/552.


zemse:
Hey @hasparus, this is definitely very compact, I was trying the example but it gives me typescript error on v4.5.4. How can it work then?

```typescript
let y: number[] & { length: 2 };
y = [1, 2];
```

```
Type 'number[]' is not assignable to type 'number[] & { length: 2; }'.
  Type 'number[]' is not assignable to type '{ length: 2; }'.
    Types of property 'length' are incompatible.
      Type 'number' is not assignable to type '2'.ts(2322)
```

contracts/v0.8.9/Issue552_Reproduction.sol
```diff
@@ -0,0 +1,40 @@
+pragma solidity ^0.8.9;
+pragma experimental ABIEncoderV2;
+
+
+library Issue552_Observer {
```
krzkaczor, line 5: Can we simplify this example? I guess something as simple as `uint256[65535] ints` triggers it already.

krzkaczor: I would propose something simple as:

```
pragma solidity ^0.8.9;
pragma experimental ABIEncoderV2;

contract Issue552_Reproduction {
  uint256[65535] public ints;

  function input(uint256[10] calldata values) public pure {}
}
```

Then we need a runtime test for `input` to ensure @zemse's comment is fixed.

krzkaczor (changes requested):

contracts/v0.8.9/Issue552_Reproduction.sol
hasparus:
```
uint256[65535] public ints;
```

☝ this doesn't emit the type, only the getter function if I see things correctly. I removed some redundant code from the reproduction though.

hasparus:
> I was trying the example but it gives me typescript error on v4.5.4. How can it work then?

I'm getting rid of that `{ length }`. It's fine on an output position, but I think it hurts DX in the end, and it's unusable as input without type assertion :/

krzkaczor:
@hasparus yeah i think that if tuple is bigger than x (10? or 6?) we should just treat it as an array.

krzkaczor (approved):

PR #560, opened 2021-11-26
Fix structName for nested array of struct

dk1a:
I think this is only relevant for ethers-v5 - I had an error with it and my function that returns Struct[][]

### DIFF
--- a/.changeset/silver-foxes-remember.md
+++ b/.changeset/silver-foxes-remember.md
@@ -1,0 +1,5 @@
+---
+"typechain": patch
+---
+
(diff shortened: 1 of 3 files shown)


krzkaczor:
@dk1a thanks! Could you remove changes to bytecodes of fixtures? not sure what happened there...

dk1a:
amended

krzkaczor:
Thanks! 🙇 

Released as: https://github.com/dethcrypto/TypeChain/releases/tag/typechain%406.0.5

PR #562, opened 2021-11-30
Show help message given `-h` flag

hasparus:
Closes https://github.com/dethcrypto/TypeChain/issues/558

_The message itself:._

```
TypeChain/packages/typechain on  cli-help [$] is 📦 v6.0.5 via  v14.17.0 
❯ node dist/cli/cli -h

  TypeChain generates TypeScript types for Ethereum contract ABIs.
  Thank you for using it!

Options

  --glob string[]                Pattern that will be used to find ABIs. Remember about adding     
                                 quotes: typechain "**/*.json", examples: ./abis/**/*.abi,
                                 ./abis/?(Oasis.abi|OasisHelper.abi).
  --target string                One of ethers-v4, ethers-v5, truffle-v4, truffle-v5, web3-v1 or   
                                 path to your custom target. Typechain will try to load package    
                                 named: @typechain/<target>, so make sure that desired package is  
                                 installed.
  --out-dir string               Output directory for generated files.
  --always-generate-overloads    Some targets won't generate unnecessary types for overloaded      
                                 functions by default, this option forces to always generate       
                                 them.
  --show-stack-traces
  -h, --help                     Prints this message.

Example Usage

  typechain --target ethers-v5 --out-dir app/contracts './contracts/*.json'


  You can read more about TypeChain at https://github.com/dethcrypto/TypeChain.
```

### DIFF
--- a/packages/typechain/test/cli/parseArgs.test.ts
+++ b/packages/typechain/test/cli/parseArgs.test.ts
@@ -1,4 +1,4 @@
-import { expect } from 'earljs'
+import { expect, mockFn } from 'earljs'
 
 import { parseArgs } from '../../src/cli/parseArgs'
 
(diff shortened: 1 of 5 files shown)


packages/typechain/test/cli/parseArgs.test.ts
```diff
@@ -62,4 +62,36 @@ describe('cli > parseArgs', () => {
...
+    })
+    process.exit = exitMock
+    const logMock = mockFn<typeof console.log>(() => {})
+    // eslint-disable-next-line no-console
+    console.log = logMock
+
+    try {
+      parseArgs()
+    } catch (err) {
+      expect((err as Error).message).toEqual('process.exit called')
```
krzkaczor, line 84: why not use: `expect().toThrow('process.exit called')`?
 
 https://earljs.dev/docs/guides/testing-errors#matching-only-part-of-an-error-message

packages/typechain/test/cli/parseArgs.test.ts
```diff
@@ -62,4 +62,36 @@ describe('cli > parseArgs', () => {
...
+    }
+
+    expect(exitMock).toHaveBeenCalledWith([0])
+    const logged = logMock.calls[0].args[0] as string
+    expect(logged.includes('Options')).toEqual(true)
```
krzkaczor, line 89: probably a manner of taste but I would personally use dedicated matchers for string ie.

```suggestion
    expect(logged).toEqual(expect.stringMatching('Options'))
```

Sidenote: we should add a validator in earl for this too ie: 

```
expect(logged).toBeAStringMatching('Options')
```
Once we merge docs generation i will for on this 😆

krzkaczor (commented):
Few questions regarding tests. Btw. i fell in love with `ts-command-line-args`. Good stuff 💪

krzkaczor (approved):

Issue #546, opened 2021-11-15
Add CLI flag to allow ignore TypeScript errors in generated code


hasparus:
We'd obviously prefer to write TypeChain in TypeScript as strong and safe as possible, but we can't guarantee that user project's config won't be stricter, revealing errors in generated code that weren't caught during the build.

I'd like to propose adding `--ts-nocheck` CLI flag and `process.env.TYPECHAIN_TS_NOCHECK` environment variable to add `/* @ts-no-check */ to generated files. This would allow users to workaround problems like https://github.com/dethcrypto/TypeChain/issues/393 on their side before they're fixed in TypeChain.

@krzkaczor what do you think?

krzkaczor:
How would you compare your solution to the one presented in: https://github.com/dethcrypto/TypeChain/pull/543 ?

hasparus:
#543 is proper solution for specific case. This issue is workaround for general case. Ideally, I'd have both.

krzkaczor:
This might be a solution to fix some of the issues (like #393) but we need to warn users that this can hide some important problems and definitely this shouldn't be the default. Also, are you positive that `/* @ts-no-check */` works inside TS files?

hasparus:
@krzkaczor

PR #564, opened 2021-11-30
Add explicit override modifier to ContractFactory subclasses

### LINKED ISSUE #393: compilation errors with TypeScript 4.3 and noImplicitOverride
TS 4.3 adds a `noImplicitOverride` compilation flag, and all methods that override a parent method are required to be marked with `override`.

@krzkaczor how should we handle this? Add a switch or detect TS version and include the override modifiers?
I don't think there is an option to disable `noImplicitOverride` selectively for the typechain-generated directories :(

Other ideas seem very ugly since they effectively disable typechecking:
* `// @ts-nocheck` at the top of the file
* `// @ts-ignore` at each meathod

hasparus:
This _should_ close #393 for Ethers v5 target.

### DIFF
--- a/.changeset/healthy-spiders-smile.md
+++ b/.changeset/healthy-spiders-smile.md
@@ -1,0 +1,5 @@
+---
+'@typechain/ethers-v5': patch
+---
+
+ContractFactory subclasses now use explicit "override" modifiers.
--- a/examples/ethers-v4/src/index.ts
+++ b/examples/ethers-v4/src/index.ts
@@ -10,7 +10,7 @@ async function main() {
   const provider = new ethers.providers.JsonRpcProvider(RPC_HOST)
   const formatEther = ethers.utils.formatEther
 
-  const dai = (new ethers.Contract(DAI_ADDRESS, abi, provider) as any) as Dai
+  const dai = new ethers.Contract(DAI_ADDRESS, abi, provider) as any as Dai
   const balance = await dai.functions.balanceOf('0x70b144972C5Ef6CB941A5379240B74239c418CD4')
 
   console.log(`Our DAI balance is: ${formatEther(balance)}`)
--- a/examples/web3-v1/src/index.ts
+++ b/examples/web3-v1/src/index.ts
@@ -10,7 +10,7 @@ async function main() {
   const web3 = new Web3(RPC_HOST)
   const fromWei = web3.utils.fromWei
 
-  const dai = (new web3.eth.Contract(abi, DAI_ADDRESS) as any) as Dai
+  const dai = new web3.eth.Contract(abi, DAI_ADDRESS) as any as Dai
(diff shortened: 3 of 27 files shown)


krzkaczor (commented):
I am a little bit worried about this PR as it's gonna break some users code due to prettier incompatibilities. Also, what do you think about doing: https://github.com/dethcrypto/TypeChain/issues/546 instead? and just slapping ts-ignore on all similar issues?

hasparus:
> as it's gonna break some users code due to prettier incompatibilities.

Don't we use that prettier version from our dependencies? I think we'd only break something if somebody has explicitly set a Prettier version in Yarn/pnpm resolutions.

krzkaczor:
> Don't we use that prettier version from our dependencies? I think we'd only break something if somebody has explicitly set a Prettier version in Yarn/pnpm resolutions.

Yeah, you're right but still, #566 seems to me like a better, more generic fix. Why make an exception for this case if we plan on ignoring all similar cases in the future. 

Furthermore, I am worried that without tests we will break it again in the near future...

hasparus:
> Yeah, you're right but still, #566 seems to me like a better, more generic fix. Why make an exception for this case if we plan on ignoring all similar cases in the future.

I think we misunderstood each other. I didn't propose #546 and create #566 as an alternative to this, I made it in addition to this PR.

#566 is more generic, but I wouldn't call it better — it's a way to work around the problem in userspace, while this PR is a specific fix on our side.

If we generate code that gets compiled and typechecked in user's projects, we should strive to generate it as strict as possible. If you plan to ignore all similar cases in the futures, we can as well make the `--ts-nocheck` flag default to `true`, as a similar case will appear every two TypeScript versions when they introduce new flag.

Could we keep this open and merge when we release a major version of TypeChain that depends on TypeScript 4.3 or higher? (`override` modifier is introduced in 4.3)

> Furthermore, I am worried that without tests we will break it again in the near future...

We have some "tests", as I added `noImplicitOverride` to our config, so our `typecheck` check tests this, but I can add a new one , with `ts-snippet` to configure TypeScript config per test perhaps?

krzkaczor:
> We have some "tests", as I added noImplicitOverride to our config,

Ahh right, this takes care of my concern about regressions. 

Okay, I didn't realize that this PR will prevent future regressions. That was my biggest concern basically -- I am fine with merging.

krzkaczor (approved):

hasparus:
We can let this lie here for a bit and mature, as it will be a breaking change for users with old TypeScript.

**What needs to be done**

- [ ] Add a memo to readme noting the required TS version per TypeChain version? [Like fp-ts does?](https://github.com/gcanti/fp-ts#typescript-compatibility)
- [ ] Mention required versions of TypeScript and Prettier in the changeset
- [ ] Ensure that the dependency version is correct

PR #573, opened 2021-12-03
Disable no-unused-import-ts ESLint rule if running inside VSCode

hasparus:
POV:

1. You import a thing.
2. You save the file.
3. The import declaration you just wrote dissapears.

---

`unused-imports/no-unused-imports-ts` is an evil rule to run inside of a text editor. I've been trying to live with it for two months, but enough is enough :P What's worse, I have autosave enabled, so I import a thing, wait a second and it dissapears :>

`VSCODE_PID` is not set when we run ESLint from the CLI so unused imports will be safely removed in `yarn test:fix`.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,3 +1,5 @@
+const inTextEditor = process.env.VSCODE_PID !== undefined
+
 /** @type {import("eslint").Linter.Config} */
 module.exports = {
   env: {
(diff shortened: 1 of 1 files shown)


krzkaczor:
Can't you deal with this in your (workspace) settings file? I use totally custom eslint config (whitelist of important rules): 

```
"eslint.options": {
    "rules": {
      "simple-import-sort/imports": "off",
      "simple-import-sort/exports": "off",
      "no-unused-vars": "off",
      "@typescript-eslint/no-unused-vars": "off",
      "no-only-tests/no-only-tests": "off",
      "unused-imports/no-unused-imports-ts": "off",
      "no-debugger": "off",
      "import/no-extraneous-dependencies": "off",
      "no-use-before-define": "off",
      "prefer-const": "off"
    }
    // "parserOptions": {
    //   "tsconfigRootDir": "absolute path"
    // }
  },
  ```

krzkaczor (approved):

PR #576, opened 2021-12-13
Multiple struct inputs with same name

### LINKED ISSUE #574: Cannot have multiple structs with the same name under different namespaces
Due to the way [`extractStructNameIfAvailable`](https://github.com/dethcrypto/TypeChain/blob/master/packages/typechain/src/parser/parseEvmType.ts#L130) is written, if you have two structs say `LibraryA.data` and `LibraryB.data`, typechain will export only one of them with the name `Data<something>`.

One solution would be to keep the names before the `.`. Another would be to keep a map of full names and add an index to each new struct encountered, so you would export `Data1`, `Data2`, etc. Happy to try one of the solutions.

zemse:
```solidity
contract MyContract
  function input_multiple_structs_with_same_name(StructsLib1.Info memory info1) public pure returns (StructsLib2.Info memory info2) {
     info2.a = address(info1.a);
     info2.b = address(info1.b);
   }
 }

 library StructsLib1 {
   struct Info {
     uint160 a;
     uint160 b;
   }
 }

 library StructsLib2 {
   struct Info {
     address a;
     address b;
   }
 }
```

Generated Typings:

```typescript
export type InfoStruct = { a: BigNumberish; b: BigNumberish };

export type InfoStructOutput = [BigNumber, BigNumber] & {
  a: BigNumber;
  b: BigNumber;
};


input_multiple_structs_with_same_name(
  info1: InfoStruct,
  overrides?: CallOverrides
): Promise<InfoStructOutput>;
```

---

Closes #574

### DIFF
--- a/.changeset/pretty-waves-leave.md
+++ b/.changeset/pretty-waves-leave.md
@@ -1,0 +1,6 @@
+---
+'@typechain/ethers-v5': major
+'typechain': major
+---
+
+Structs will be namespaced using contract name when available
--- a/contracts/v0.6.4/DataTypesInput.sol
+++ b/contracts/v0.6.4/DataTypesInput.sol
@@ -70,4 +70,23 @@ contract DataTypesInput {
   function input_struct2_tuple(Struct2[3] memory input) public pure returns (Struct2[3] memory) {
     return input;
   }
+
+  function input_multiple_structs_with_same_name(StructsLib1.Info memory info1) public pure returns (StructsLib2.Info memory info2) {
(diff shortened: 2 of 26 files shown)


krzkaczor:
I assume this is related to: https://github.com/dethcrypto/TypeChain/issues/574

krzkaczor (approved):

MHarris021:
This PR causes a breaking change that prevents generics or type inheritance for contracts due to the contractName type being set to a hard coded string instead of just a string type.

Issue #581, opened 2021-12-24
Generate tagged unions for contract types


ShravanSunder:
If typechain can generate tagged unions for the contract types and factories this would make manipulation and sorting for libraries a lot easier.

For example in the definition of the following contracts
```
export class YourContract extends BaseContract {
  contractName: 'YourContract';
  connect(signerOrProvider: Signer | Provider | string): this;
...
```
```
export class Dai17 extends BaseContract {
  contractName: 'Dai17';
  connect(signerOrProvider: Signer | Provider | string): this;
...
```

I can use union discrimination and other typescript properties to properly type contracts that are stored in a library.  Such as a contract context in react.  

Related to #579  in that it allows for easy manipulation of types

ShravanSunder:
perhaps related to #577

krzkaczor:
closed with https://github.com/dethcrypto/TypeChain/pull/648

Issue #579, opened 2021-12-19
Can TypeChain generate a list of contracts in addition to definitions


ShravanSunder:
I'm developing `scaffold-eth-typescript` and `eth-hooks`.  I was wondering if typechain could generate a index.ts file that would generate a record of contracts to factories as an index file?

for example if your contracts include `contract1.sol`, yourcontract.sol', 'contract2.sol', typechain would also generate an `index.ts` file
```
export const definitions = {
'contract1`: contract1__factory
`'yourcontract`: yourcontract__factory,
 ...
}
```

it would help to create a development environment that would autoload contracts with types.

krzkaczor:
Hey. Yeah we definitely can look into how we can make typechain integrate better with other tools. Have you looked into [eth-sdk](https://github.com/dethcrypto/eth-sdk)?

ShravanSunder:
@krzkaczor yes!  i'm using eth-sdk as well and they way it outputs indexes are extremely useful.

However, some of the contracts are locally generated.  Some are external contracts.  
- For the external contracts I'm using eth-sdk
- for local contracts that will be built and tested with hardhat, I am using typechain to generate everything.  However for typing and more automatic in the library, it would be great if an index.ts was generated with the rest of the files (definitions and factories)

ShravanSunder:
@krzkaczor  if you can let me know which file to look at I can contribute =)

Issue #577, opened 2021-12-15
[Feature request] Any plans to add type guards?


davidlee1435:
Are there any plans to automatically generate [type guards](https://dev.to/ricardopaul/typescript-type-guards-simply-explained-oik)?

krzkaczor:
Type guards for what exactly?

davidlee1435:
Type guards for contracts. I'd like a way to check whether an `ethers.Contract` instance can be cast to a type generated by TypeChain. Given a contract `SomeContract`, I'm thinking it could look like this:

```
const isSomeContract = (contract: any): boolean => { 
  // code that returns true if the input contains all functions 
  //defined in the SomeContract generated type
}
```

If not, is there a recommended way to check whether some arbitrary variable can be cast to a type?

PR #582, opened 2021-12-24
#581 descriminated-unions support

ShravanSunder:
added contract.name to generated contract, interface and factory
allows for type manipulation and conditional types

caused by:
- #581 

related to:
- #577 

# #581 

If typechain can generate tagged unions for the contract types and factories this would make manipulation and sorting for libraries a lot easier.

For example in the definition of the following contracts
```
export class YourContract extends BaseContract {
  contractName: 'YourContract';
  connect(signerOrProvider: Signer | Provider | string): this;
...
```
```
export class Dai17 extends BaseContract {
  contractName: 'Dai17';
  connect(signerOrProvider: Signer | Provider | string): this;
...
```

I can use union discrimination and other typescript properties to properly type contracts that are stored in a library.  Such as a contract context in react.  

Related to #579  in that it allows for easy manipulation of types


.gitignore
```diff
@@ -28,3 +28,4 @@ packages/hardhat-test/artifacts
 examples/*/yarn.lock
 yarn-error.log
 packages/test-e2e/test/*/out
+.DS_Store
```
krzkaczor, line 31: Please move this to your global git ignore filee

krzkaczor (changes requested):
Interesting change. "Artificial" fields like this, will require casting via `any/unknown` but I think it's required anyway currently. Can you run `yarn test:fix` to regenerate all generated files and commit them? (https://github.com/dethcrypto/TypeChain/blob/master/CONTRIBUTING.md)

.gitignore
ShravanSunder: done

ShravanSunder:
> Interesting change. "Artificial" fields like this, will require casting via `any/unknown` but I think it's required anyway currently. Can you run `yarn test:fix` to regenerate all generated files and commit them? (https://github.com/dethcrypto/TypeChain/blob/master/CONTRIBUTING.md)

Sorry about the delay.

@krzkaczor  I ran yarn `test:fix` however there doesn't seem to be any files  that I changed that I could commit.

image

ShravanSunder:
@krzkaczor  hi, are things ok with this PR to merge?

.gitignore
```diff
@@ -27,4 +27,4 @@ packages/hardhat-test/cache
 packages/hardhat-test/artifacts
 examples/*/yarn.lock
 yarn-error.log
-packages/test-e2e/test/*/out
+packages/test-e2e/test/*/out
```
krzkaczor, line 30: can you revert this change?

krzkaczor (commented):
You need to commit changed files after running `yarn test` we treat these files as snapshot tests.

ShravanSunder:
> You need to commit changed files after running `yarn test` we treat these files as snapshot tests.

hi @krzkaczor  I just ran yarn test and pushed the changes.  I also fixed gitignore ( I think master has that line included)

krzkaczor:
@ShravanSunder thanks for working on this. Can you add me on telegram: `krzkaczor` i wanted to dive into some details.

MHarris021:
> Interesting change. "Artificial" fields like this, will require casting via `any/unknown` but I think it's required anyway currently. Can you run `yarn test:fix` to regenerate all generated files and commit them? (https://github.com/dethcrypto/TypeChain/blob/master/CONTRIBUTING.md)

This merge breaks the ability to use generic types for Contracts as the contractName field type is now a hard-coded string. 

Previously an ERC20 that implemented IERC20 could be used in a function that declared <T extends IERC20>, i.e... 

`export async function getTokensTotalSupplies(tokens: IERC20Metadata[]) {
  tokenNumberGuard(tokens);
  return Promise.all(tokens.map(token => token.totalSupply()));
}`

would work for any Contract that could be Typecast into an @openzeppelin/contracts/token/ERC20/extensions/IERC20Metadata.sol . Now it doesn't work.

@ShravanSunder can you help me understand what additional functionality the contractName field provides to you that you couldn't get by doing a simple typeof contract or type = ContractType1|ContractType2|ContractType3 ?

krzkaczor:
@MHarris021 thanks for this report. 

@MHarris021 @ShravanSunder do you see any other option then hiding this feature behind a flag?

MHarris021:
@ShravanSunder it appears that you are using the contractName field on the interface here: [https://github.com/scaffold-eth/eth-hooks/blob/next/src/models/contractContextTypes.ts](https://github.com/scaffold-eth/eth-hooks/blob/next/src/models/contractContextTypes.ts) ? Is that correct?

MHarris021:
@krzkaczor I created a generic test case not using Typechain and adding `string` to the contractName type seemed to work fine.

`export interface Stuff {
  stuffField: string|"Stuff";
  getName():string;
}

export interface AwesomeStuff {
  stuffField: string|"AwesomeStuff";
  getName():string;
  getAwesomeName():string;
}

export function doStuff<T extends Stuff>(stuff:T):string{
  return stuff.getName();
}

class AwesomeStuffImpl implements AwesomeStuff{

  stuffField: string | "AwesomeStuff";
  private readonly name: string;

  constructor() {
    this.stuffField="AwesomeStuff";
    this.name="StuffImpl";
  }
  getAwesomeName(): string {
    return `Awesome${this.getName()}`;
  }

  getName(): string {
    return this.name;
  }

}

const stuff = new AwesomeStuffImpl();

doStuff(stuff);
` 

I'd be happy to address this if you want me to.

ShravanSunder:
> > Interesting change. "Artificial" fields like this, will require casting via `any/unknown` but I think it's required anyway currently. Can you run `yarn test:fix` to regenerate all generated files and commit them? (https://github.com/dethcrypto/TypeChain/blob/master/CONTRIBUTING.md)
> 
> This merge breaks the ability to use generic types for Contracts as the contractName field type is now a hard-coded string.
> 
> Previously an ERC20 that implemented IERC20 could be used in a function that declared , i.e...
> 
> `export async function getTokensTotalSupplies(tokens: IERC20Metadata[]) { tokenNumberGuard(tokens); return Promise.all(tokens.map(token => token.totalSupply())); }`
> 
> would work for any Contract that could be Typecast into an @openzeppelin/contracts/token/ERC20/extensions/IERC20Metadata.sol . Now it doesn't work.
> 
> @ShravanSunder can you help me understand what additional functionality the contractName field provides to you that you couldn't get by doing a simple typeof contract or type = ContractType1|ContractType2|ContractType3 ?
@MHarris021 
The reason for this is to allow for strict union discrimination when you have a Base type.  For example
```
interface FishContract extends BaseContract {
contractName : 'Fish'
}

interface MonkeyContract extends BaseContract {
contractName : 'Monkey'
}

const func = <(unknownContract: BaseContract & {contractName: string}) => {
  if (unknownContract.contractName === 'Monkey') {
     // this contract is now typed all the types can be used
  }
}

```

ShravanSunder:
> @krzkaczor I created a generic test case not using Typechain and adding `string` to the contractName type seemed to work fine.
> 
> ```
> export interface Stuff {
>   stuffField: string|"Stuff";
>   getName():string;
> }
> 
> export interface AwesomeStuff {
>   stuffField: string|"AwesomeStuff";
>   getName():string;
>   getAwesomeName():string;
> }
> 
> export function doStuff<T extends Stuff>(stuff:T):string{
>   return stuff.getName();
> }
> 
> class AwesomeStuffImpl implements AwesomeStuff{
> 
>   stuffField: string | "AwesomeStuff";
>   private readonly name: string;
> 
>   constructor() {
>     this.stuffField="AwesomeStuff";
>     this.name="StuffImpl";
>   }
>   getAwesomeName(): string {
>     return `Awesome${this.getName()}`;
>   }
> 
>   getName(): string {
>     return this.name;
>   }
> 
> }
> 
> const stuff = new AwesomeStuffImpl();
> 
> doStuff(stuff); 
> ```
> 
> I'd be happy to address this if you want me to.

as I mentioned in #603 this change does not work as explained in the above post

@MHarris021  @krzkaczor  we could make it an option in typechain.  I think that works for me.

ShravanSunder:
@MHarris021  as a temporary workaround.
you can cast into IERC20Metadata:
export async function getTokensTotalSupplies(tokens: IERC20Metadata[]) { tokenNumberGuard(tokens); return Promise.all(tokens.map(token => token.totalSupply())); }

if you're calling this function.  if IERC20Metadata is based on typechain.  
https://www.typescriptlang.org/docs/handbook/utility-types.html#omittype-keys
```
getTokensTotalSupplies(tokens: Omit<IERC20Metadata, 'contractName'>[]) {
...
}
```

Lumyo:
@ShravanSunder I tried the suggested workaround, but it doesnt work and I'm still getting erros, because there are a bunch of nested `contractName` fields which arent ommited by this (Ex : `interface.contractName`, `deployed().contractName`, `connect().contractName`, ...)

ShravanSunder:
> @ShravanSunder I tried the suggested workaround, but it doesnt work and I'm still getting erros, because there are a bunch of nested `contractName` fields which arent ommited by this (Ex : `interface.contractName`, `deployed().contractName`, `connect().contractName`, ...)

If you share your branch in your fork I can try to take a look this weekend

PR #587, opened 2021-12-29
Add missing parameter names to event filter

hasparus:
Closes https://github.com/dethcrypto/TypeChain/issues/575

### DIFF
--- a/packages/target-ethers-v5/test/generation.test.ts
+++ b/packages/target-ethers-v5/test/generation.test.ts
@@ -66,7 +66,7 @@ describe(generateEventFilters.name, () => {
     expect(actual).toEqual(expect.stringMatching('arg1?: null'))
   })
 
-  it('generates event filters', () => {
+  it('generates argument names from event field names in the ABI', () => {
     const abi = ([
       {
         anonymous: false,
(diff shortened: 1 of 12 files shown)


SoheilAdmin:
bc1quc57lgsdhrxurmnp3hre07zfpx04jepdmcl65u

krzkaczor (approved):

packages/target-ethers-v5/test/generation.test.ts
```diff
@@ -29,3 +30,122 @@ describe('Ethers generation edge cases', () => {
...
+        ],
+        name: 'summoned',
+        type: 'event',
+      },
+    ] as any) as RawAbiDefinition[]
+
+    const contract = parse(abi, 'Rarity')
+    const actual = Object.values(contract.events).map(generateEventFilters)
+
+    const expected = [
```
krzkaczor, line 126: i am not sure if we need such tests as we treat generated typings as a sort of snapshot tests.

hasparus: I added this one because it's easier to add `it.only` inside of Mocha test to work on a localized problem and the test descriptions document intention.

I can remove it if you'd prefer.

krzkaczor: I think it will easily break once we change anything. I would rather delete it as it doesn't provide any additional value to the snapshot test.

PR #588, opened 2021-12-29
Fixes issue with signatures breaking if overriden with arrays.

krzkaczor:
Closes: https://github.com/dethcrypto/TypeChain/issues/428


hasparus:
The PR looks nice. Generated type names are hideous, but I don't think we can do anything about it :>

hasparus (approved):

hasparus:
The CI fails exactly on the problem that's solved in this one, right?

image

It seems that we're getting an old version of the package installed.

krzkaczor:
@hasparus yeah, I didn't noticed it earlier. I tried restarting CI pipeline too - no success. Needs investigation.

krzkaczor:
@hasparus got 'em. Problem was missing, explicit dependency on typechain in test packages: https://github.com/dethcrypto/TypeChain/pull/588/files#diff-846c9f270bd17800449a0a8b5116663eedf2e7f37d041ad687e99b570b742369R20

For some reason it works just fine on linux but not on windows 🤔

PR #600, opened 2022-01-12
update internal task name to match hardhat convention

ItsNickBarry:
The Hardhat internal tasks use colons rather than hyphens to simulate task namespaces: https://github.com/nomiclabs/hardhat/blob/master/packages/hardhat-core/src/builtin-tasks/task-names.ts.  TypeChain should do the same.  Change should not be considered breaking because the proper way to reference a task name is by importing `constants.ts`.

### DIFF
--- a/.changeset/rich-penguins-shave.md
+++ b/.changeset/rich-penguins-shave.md
@@ -1,0 +1,5 @@
+---
+"@typechain/hardhat": major
+---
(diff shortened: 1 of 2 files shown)


krzkaczor:
Thanks @ItsNickBarry! It's gonna be released in the next major.

PR #603, opened 2022-01-14
Fixes Issue-601 and still preserves changes made by PR-582 to support…

### LINKED ISSUE #601: Merge of #582 breaks the ability to use generic types for Contracts as the contractName field type is now a hard-coded string.
Merge of #582 breaks the ability to use generic types for Contracts as the contractName field type is now a hard-coded string.

Previously an ERC20 that implemented IERC20 could be used in a function that declared , i.e...

```
export async function getTokensTotalSupplies(tokens: IERC20Metadata[]) { 
  tokenNumberGuard(tokens);
  return Promise.all(tokens.map(token => token.totalSupply())); 
}
```

would work for any Contract that could be Typecast into an @openzeppelin/contracts/token/ERC20/extensions/IERC20Metadata.sol . Now it doesn't work.

changing the contractName field to include string as a possible type seems to fix the issue

MHarris021:
… tagged unions for contract types.

Fixes #601 

Added several OZ basic contracts to use for testing purposes.
Added .idea folder to the .gitignore file to ignore JetBrains based IDE config files from being stored in git.
Created Generics.test.ts to verify the ability to support type generics.
Modified common.ts in the packages/target-ethers-v5-test directory to support deploying contracts with arguments and to create a basic function that uses generics.


MHarris021:
@ShravanSunder can you please verify that my changes didn't break your requirements?

ShravanSunder:
Will check today

ShravanSunder:
It does break my requirements. The whole point of specific types is type dewceimination in typescript. It is not meant to be used by javascript code.

krzkaczor:
Thanks for working on this @MHarris021.

I think that we need to revert change introduced in https://github.com/dethcrypto/TypeChain/pull/582 and hide them under optional flag. Do we all agree that this is the way forward? @ShravanSunder @MHarris021 ? 

@MHarris021 do you want to work on this? (in this PR perhaps)

ShravanSunder:
> Thanks for working on this @MHarris021.
> 
> I think that we need to revert change introduced in #582 and hide them under optional flag. Do we all agree that this is the way forward? @ShravanSunder @MHarris021 ?
> 
> @MHarris021 do you want to work on this? (in this PR perhaps)
@krzkaczor 
Options flag would be best!  Is it possible to do the revert and options flag in the same PR, the way it doesn’t break some of my new changes to eth-hooks.

krzkaczor:
@ShravanSunder I would prefer to release it as the next major. And just leave 7.x branch as it is.

MHarris021:
@ShravanSunder thank you for the example. There may be some nuances of Typescript that are eluding me in what you are trying to do. I was not aware that you could omit specific fields when using Generics.

@krzkaczor you are welcome. I'm happy to help. To make sure I understand correctly, my understanding is that you would appreciate my help adding in the flag/option to add a contractName to the contract types. Do you prefer that the default behavior is to add the field or not have it present? And, do you want me to add `string` to the field's type or just not have it appear? Also, I realized that I needed to change `'` to `"` to be compatible with windows platforms for my solidity contracts. I'll make sure to add the new contract types that I used for testing to git. Didn't realize that i made that mistake. I've since enabled Github Actions on my forked repo, so I'll check the CI output before resubmitting a PR. Once I know which way you prefer the option to work.

krzkaczor:
@MHarris021 awesome!

> To make sure I understand correctly, my understanding is that you would appreciate my help adding in the flag/option to add a contractName to the contract types. 

Yes, exactly.

>  Do you prefer that the default behavior is to add the field or not have it present?

I think that this whole feature of generating `contractName` should be hidden behind the flag because it's not needed usually. 

> And, do you want me to add string to the field's type or just not have it appear?

I think that it needs to use only literal strings so no ` | string`.

>  Didn't realize that i made that mistake. I've since enabled Github Actions on my forked repo, so I'll check the CI output before resubmitting a PR. Once I know which way you prefer the option to work.

Yeah, no worries. Our repo setup is quite custom. 

Thanks for your help!

MHarris021:
@krzkaczor i'll let you know when i'm ready for you to retest this PR. I'm having problems getting my relative imports to work correctly for the Windows CI Workflow on the v0.8.9 solc contracts. I think it may be due to the way the base path is interpreted. I'll do some testing tomorrow to verify.

ShravanSunder:
@MHarris021 @krzkaczor So i was looking at my usecases for discriminated union.

- What was most important was the `ContractFactory` to have the contractName to descriminate them from a record. 
-  The created contracts could then infered based on the `Record of ContractFactory`

I could test this out and we could remove contractName from Contracts and ethers.utils.interfaces that typechain generates.  This would avoid the runtime issues form above but create Factories with typed connectors.

Let me know if this is ok, and I can test this out.  if this works, I can make a PR this weekend

nkuba:
Hey guys, can we expect the fix soon? I'm really looking forward to it. 🤞🏻

krzkaczor:
Thanks, @MHarris021 for the initial work on this. I took your idea for a regression test and moved it to https://github.com/dethcrypto/TypeChain/pull/648. The plan is to introduce a special flag for ppl interested in such a feature. By default it's off.

We should release it in next major -- next week.

PR #607, opened 2022-01-20
Use `import type` for types instead of `import`

DeepDoge:
Made it generate type imports with `import type` instead of  `import` alone.

**What I changed:**
- Before `createImportDeclaration` and `createImportTypeDeclaration` functions were both doing the exact same thing, now `createImportTypeDeclaration` imports with `import type`.
- When `createImportsForUsedIdentifiers` used, if the `moduleSpecifier` is starting with `type`, it's using `createImportTypeDeclaration` for it instead of `createImportDeclaration`.

I did this because this is how it's suppose to be and it was causing problems with Svelte's Vite.


packages/target-ethers-v5/src/codegen/index.ts
```diff
@@ -189,17 +189,9 @@ export function codegenContractFactory(contract: Contract, abi: any, bytecode?:
...
-        'utils',
-        'Contract',
-        'ContractFactory',
-        'PayableOverrides',
-        'Overrides',
-        'BytesLike',
-        'BigNumberish',
-      ],
-      '@ethersproject/providers': ['Provider', 'TransactionRequest'],
+      ethers: ['Signer', 'utils', 'Contract', 'ContractFactory', 'PayableOverrides', 'BytesLike', 'BigNumberish'],
```
krzkaczor, line 192: I can change this line to:
```suggestion
      'type ethers': ['Signer', 'utils', 'Contract', 'ContractFactory', 'PayableOverrides', 'BytesLike', 'BigNumberish'],
```
(merge with next line) and all tests still pass.

krzkaczor (changes requested):
Personally, I don't use `import/export type` that much but I get that some tooling might require it. The problem with this PR is that right now our tests can't catch any mistakes related to this. I can change `import`  to `import type` and the other way around and CI still passes. 

How can we verify that current changes work?

packages/target-ethers-v5/src/codegen/index.ts
DeepDoge: Do you also want me to edit tests so when something that is a type is imported without `import type` or the other way around, we catch it in the tests?

DeepDoge:
Do you also want me to edit tests so when something that is a type imported without import type or the other way around, we catch it in the tests?

DeepDoge:
- Added `"importsNotUsedAsValues": "error"` in tsconfig in the root. 
-- This should make the `typecheck` fail if there is any type imported without `import type` and haven't been used as value inside the file.
- And also updated all of the files in the project accordingly.

scripts/.eslintrc.js
```diff
@@ -0,0 +1,6 @@
+/** @type {import("eslint").Linter.Config} */
```
krzkaczor, line 1: Is this file needed?

tsconfig.json
```diff
@@ -9,7 +9,8 @@
     "sourceMap": true,
     "composite": true,
     "incremental": true,
-    "exactOptionalPropertyTypes": true
+    "exactOptionalPropertyTypes": true,
+    "importsNotUsedAsValues": "error"
```
krzkaczor, line 13: what if we enforce this only in `packages/target-ethers-v5/tsconfig.json`? I don't feel that it's necessary to use accross the whole codebase.

krzkaczor (changes requested):

krzkaczor:
@DeepDoge thanks! This flag `importsNotUsedAsValues` is what we need (i think). It just should be scoped to ethers test package. I hope there is automatic way to fix it 😅

DeepDoge:
@krzkaczor I mean, I believe its good that if we use for the whole codebase. It's more consistent and I already updated the whole codebase for it.
Why it would be good because for example while im running the tests, i also realized that other than ethers v5, ethers v4 and web3 had the same problem too. And I fixed those too.
I didn't realized those before because i was only using ethers5 in my project.

But if you really don't want it for the whole codebase, I can only scope it for, ethers v5, v4 and web3.

scripts/.eslintrc.js
DeepDoge: I realized that it was giving ide errors because the folder had files with console.logs but the root eslint config doesn't allow console.logs.

But scripts folder contains files for cli scripts.

So i suppose it's needed.

krzkaczor:
> But if you really don't want it for the whole codebase, I can only scope it for, ethers v5, v4 and web3 tests.

Yes, I would prefer to do it like this. I am not sure if there is an automatic way to enforce proper use of `import type` and I don't want to put the additional burden on maintainers/contributors.

scripts/.eslintrc.js
krzkaczor: Hmm but this way this file is also not used by anything.... I think the real solution would be to enforce root eslint config for every single file (right now it's only for code that is actually published) and then ignore valid console.logs with eslint ignore comments. I'd remove this from this PR and do this in a separate one.

DeepDoge:
@krzkaczor can you check now?

packages/hardhat-test/test/hardhat.test.ts
```diff
@@ -20,8 +20,12 @@ describe('TypeChain x Hardhat', () => {
     const artifact = require('../artifacts/contracts/Counter.sol/Counter.json') as { abi: Interface; bytecode: string }
 
     const counterFactory1 = new Counter__factory()
+    counterFactory1
```
hasparus, line 23: Why is this needed?

krzkaczor:
Superseded by https://github.com/dethcrypto/TypeChain/pull/636

krzkaczor:
Thanks, @DeepDoge for your work on this 🙇  I merged it as #636.

PR #629, opened 2022-03-01
Fix sample tsconfig.json excerpt in README.md

devanonon:
Since newer versions of typechain default to `./typechain-types` for the type outputs I think the README should match the default as it easily confuses new users.


krzkaczor:
@devanonon thanks! Merged 🔥

PR #636, opened 2022-03-07
Use 'import type' for types instead of 'import'

krzkaczor:
- importing types without 'import type' was problamatic with svelte vite


krzkaczor:
Original work done in: https://github.com/dethcrypto/TypeChain/pull/607

hasparus (approved):

PR #642, opened 2022-03-09
Fix interface overloads in ethers-v5 target

chmanie:
Nothing fancy, here are the changes:

The methods 

```
getEvent
getFunction
decodeFunctionResult
encodeFunctionData
```

on the `ContractiInterface` of the `ethers-v5` target now follow the following rules:

- If these entities are not overloaded in the contract ABI and `--always-generate-overloads` is off, just the entities' names are used (without the signature)
- If the entities are overloaded, only signatures are used to disambiguate them
- If `--always-generate-overloads` is on, additional overloads are generated for functions that are not ambiguous

For the method and event names in the `events` and `functions` properties _only_ signature names are used (that's according to the ethers API - they don't use shorthand naming).

### DIFF
--- a/packages/target-ethers-v5/src/codegen/events.ts
+++ b/packages/target-ethers-v5/src/codegen/events.ts
@@ -74,8 +74,8 @@ export function generateEventArgType(eventArg: EventArgDeclaration): string {
   return eventArg.isIndexed ? `${generateInputType({ useStructs: true }, eventArg.type)} | null` : 'null'
 }
 
-export function generateGetEventOverload(event: EventDeclaration): string {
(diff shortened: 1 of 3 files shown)


krzkaczor:
@chmanie we commit modified files back to the git (we treat them as snapshots), you didn't commit them so that's why CI is failing.

chmanie:
Okay, I'm sorry but I don't know what to do here.

I did `yarn test:fix` and added the changed files. Anything else?

krzkaczor:
Thanks for this! I think the problem is that you didn't rebuild TypeChain with `yarn build` and hence there were some unrelated changes in the output files. 

Superseded by: https://github.com/dethcrypto/TypeChain/pull/643

PR #643, opened 2022-03-10
Fix overloads in ethers-v5 target

krzkaczor:
https://github.com/dethcrypto/TypeChain/pull/642


krzkaczor:
@hasparus to rewrite it to use union types as it's more flexible that way.

Issue #605, opened 2022-01-20
ethers-v5 at v8 and higher target not properly generating typing for a `Struct[][]`


prestwich:
Solidity:
```
    function executeGovernanceActions(
        GovernanceMessage.Call[] calldata _localCalls,
        uint32[] calldata _domains,
        GovernanceMessage.Call[][] calldata _remoteCalls
    ) external
```

correct output generated in `@typechain/ethers-v5@7.2.0`:
```
    executeGovernanceActions(
      _localCalls: { to: BytesLike; data: BytesLike }[],
      _domains: BigNumberish[],
      _remoteCalls: { to: BytesLike; data: BytesLike }[][],
      overrides?: Overrides & { from?: string | Promise<string> }
    ): Promise<ContractTransaction>;
```

incorrect output generated in `@typechain/ethers-v5@8.0.5`:
```
executeGovernanceActions(
  _localCalls: CallStruct[], 
  _domains: BigNumberish[], 
  _remoteCalls: CallStruct[],   // <---- error is here. This should be CallStruct[][]
  overrides?: Overrides & { from?: string | Promise<string>; }
): Promise<ContractTransaction>;
```

v9 has the same problem, but puts the struct under a namespace

krzkaczor:
This works in the latest relase.

PR #647, opened 2022-03-11
Fix ethers v5 target incorrect type generated for struct array array

zemse:
EDIT by Kris: 
Tracked via: DET-40

#605

### DIFF
--- a/packages/target-ethers-v5/src/codegen/types.ts
+++ b/packages/target-ethers-v5/src/codegen/types.ts
@@ -49,7 +49,7 @@ export function generateInputType(options: GenerateTypeOptions, evmType: EvmType
           evmType.structName
             ? evmType.structName.toString() + STRUCT_INPUT_POSTFIX
             : generateInputType({ ...options, useStructs: true }, evmType.itemType),
-          evmType.size,
+          evmType.originalType,
         )
       }
     case 'boolean':
@@ -88,7 +88,7 @@ export function generateOutputType(options: GenerateTypeOptions, evmType: EvmOut
         evmType.structName
           ? evmType.structName.toString() + STRUCT_OUTPUT_POSTFIX
           : generateOutputType({ ...options, useStructs: true }, evmType.itemType),
-        evmType.size,
+        evmType.originalType,
       )
     case 'boolean':
       return 'boolean'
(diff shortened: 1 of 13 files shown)


krzkaczor:
I think this implementation is acceptable but we should consider parsing nested types in the TypeChain core, to make targets simpler.

packages/target-ethers-v5/src/codegen/types.ts
```diff
@@ -141,10 +141,22 @@ export function generateOutputComplexTypesAsObject(
   return namedElementsCode
 }
 
-function generateArrayOrTupleType(item: string, length?: number) {
-  if (length !== undefined && length < 6) {
-    return `[${Array(length).fill(item).join(', ')}]`
-  } else {
-    return `(${item})[]`
+function generateArrayOrTupleType(item: string, originalType: string): string {
```
krzkaczor, line 144: I think all of this should be moved to `parseEvmType` directly in TypeChain core.

krzkaczor (changes requested):

krzkaczor:
Also take a look at the implementation in truffle target b/c it's very clean and recursive.

zemse:
I've refactored the code and the existing tests pass. As a summary of the changes:

- In the typechain package, struct types without structName were being registered. So the `fix` commit [7289bbd](https://github.com/dethcrypto/TypeChain/pull/647/commits/7289bbdb9c6faed74d4afb0e34d4dfdaa38c3487) simply adds a check in registerStruct function (inside abiParser file) to skip such struct from registering.
- Struct with "array" type was being registered (when chronologically the array type was being encountered in the ABI), the same "fix" commit adds logic in registerStruct function to go deep and register the "tuple" if an "array" is encountered. Also a problem with this was, `structName` was not set, hence to make this work, "internalType" arg is passed recursively in `parseEvmType` function which sets the `structName` prop for the struct tuple too.
- Fixes ethers v5 test case in struct.test.ts for CallStruct which expects it to be array type. [link](https://github.com/dethcrypto/TypeChain/blob/master/packages/target-ethers-v5/test/structs.test.ts#L53). Here CallStruct is being expected as array.

Finally, the "refactor" commit is able to implement the normal recursive logic which looks cleaner (due to changes in previous commit).

Since these are breaking changes, it would be great if you can have a good enough look (I feel very less confident because I just made changes and it worked in first try lol). I just thought about a test case, will add it right now, just to be sure.

hasparus (approved):

PR #650, opened 2022-03-14
Migrate to PNPM and make `tryRequire` work with symlinks

hasparus:
There are two changes here:

- We're moving to PNPM to make builds more predictable and faster.
- As our `tryRequire` in `typechain` did not work with symlinks too well, I've added another `require.resolve` case.


README.md
```diff
@@ -110,7 +110,7 @@ you will get TypeScript typings compatible with a given library.
 
 ### Step by step guide
 
-Install TypeChain with `yarn add --dev typechain` and install desired target.
+Install TypeChain with `pnpm install --save-dev typechain` and install desired target.
```
krzkaczor, line 113: I'd keep yarn here, users don't need to know that we use pnpm

package.json
```diff
@@ -47,8 +38,11 @@
     "eslint-plugin-unused-imports": "^1.1.1",
     "mocha": "^8.2.0",
     "prettier": "^2.3.1",
-    "ts-node": "^8.5.4",
+    "ts-node": "^10.7.0",
     "typescript": "^4.6",
     "wsrun": "^5.2.4"
```
krzkaczor, line 43: we can drop wsrun, right?

packages/hardhat-test/package.json
```diff
@@ -11,16 +11,18 @@
...
     "@nomiclabs/hardhat-ethers": "^2.0.2",
-    "@typechain/ethers-v5": "*",
-    "@typechain/hardhat": "*",
+    "@typechain/ethers-v5": "workspace:^9.0.0",
+    "@typechain/hardhat": "workspace:^5.0.0",
     "ethers": "^5.4.7",
     "hardhat": "^2.0.10",
     "test-utils": "1.0.0",
-    "typechain": "*"
+    "typechain": "workspace:^7.0.1"
```
krzkaczor, line 26: what if we bump major? i think all of these, local dependencies should use *any* TC version that is available locally.

packages/hardhat-test/tsconfig.types.json
```diff
@@ -1,9 +1,9 @@
 {
   "extends": "../../tsconfig.json",
   "include": ["./typechain-types"],
-  "exclude": ["node_modules"],
   "compilerOptions": {
     "esModuleInterop": true,
-    "lib": ["ES2018", "DOM"]
+    "lib": ["ES2018", "DOM"],
+    "skipLibCheck": true
```
krzkaczor, line 7: why this?

pnpm-workspace.yaml
```diff
@@ -0,0 +1,6 @@
+packages:
+  # all packages in subdirs of packages/ and components/
+  - 'packages/**'
+  - 'components/**'
```
krzkaczor, line 4: what's this?

README.md
hasparus: True. This might be a separate matter,.

What package manager should we refer to in our user-facing readmes?
Should we just use `npm` here as it's the _"default"_?

packages/hardhat-test/tsconfig.types.json
hasparus: There are some errors in `buffer` types clashing with our deps, and we don't really care about them in `hardhat-test`.

I'd like to move towards faster testing pipeline to boost productivity in this repo.

pnpm-workspace.yaml
hasparus: A blooper, sorry for that.

README.md
krzkaczor: yeah maybe we should rewrite it to simply use `npm`

scripts/compile-contracts.ts
```diff
@@ -22,22 +19,6 @@ function main() {
 
 main()
 
-function panicOnNpm7() {
```
krzkaczor, line 25: :+1:

krzkaczor (approved):

packages/hardhat/src/config.ts
```diff
@@ -1,8 +1,8 @@
-import { HardhatUserConfig } from 'hardhat/types'
+import { HardhatConfig } from 'hardhat/types'
 
 import { TypechainConfig } from './types'
 
-export function getDefaultTypechainConfig(config: HardhatUserConfig): TypechainConfig {
+export function getDefaultTypechainConfig(config: HardhatConfig): TypechainConfig {
```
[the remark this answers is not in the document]
hasparus: Yeah,

1. We had a type error there that showed up after we stopped hoisting node_modules :D
2. We get the config object from `'hardhat/config'::extendConfig`, so it's after parsing user config.

packages/target-web3-v1-test/test/Events.test.ts
```diff
@@ -102,8 +107,8 @@ describe('Events', () => {
       await contract.methods.emit_event1().send({
         from: accounts[0],
         gas: GAS_LIMIT_STANDARD,
-        maxFeePerGas: new BigNumber('1'),
-        maxPriorityFeePerGas: new BigNumber('1'),
+        maxFeePerGas: new BigNumber(10 ** 9),
```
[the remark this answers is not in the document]
hasparus: I updated Ganache — the last version is marked as deprecated on NPM what leads to a very nasty warning during our pnpm install, and the Ganache v7 requires fees to be above ~0.7 gwei.

scripts/compile-contracts.ts
hasparus: Finally!

PR #654, opened 2022-03-17
Values name clash

### LINKED ISSUE #623: Event argument name conflicts with JS method values()
Hi all,

I've just noticed that Openzeppelin's IGovernor.sol interface defines an event with argumetn `values` (https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/governance/IGovernor.sol#L28-L37)

When I try to use the generated typings, I get that `event.args.values` is `undefined` - I believe this is because of the clash between the argument name and the JS object method `values()`

Is there a way to check for reserved keywords and pre/su-fix the name `values` with an underscore or something like that?

I can access it by index as well, but it would be messy to mix accesing content by `.` and `[]`. It might be also an issue for other reserved words.

hasparus:
Work in progress. Based on #652, so it will need a rebase.

Will close #623, when it's ready.


hasparus:
I'm a bit conflicted about this change. @krzkaczor can you take a look and tell me if you think it's even worth it? Do you see any alternative solutions?

PR #655, opened 2022-03-18
Crash Ethers V5 target if TS version < 4.3

hasparus:
Looks more or less like this:

### DIFF
--- a/packages/target-ethers-v5/src/index.ts
+++ b/packages/target-ethers-v5/src/index.ts
@@ -45,6 +45,8 @@ export default class Ethers extends TypeChainTarget {
   constructor(config: Config) {
     super(config)
 
+    panicOnOldTypeScriptVersion()
+
     const { cwd, outDir, allFiles } = config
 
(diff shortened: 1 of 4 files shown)


packages/target-ethers-v5/src/index.ts
```diff
@@ -195,3 +197,39 @@ function createRootIndexContent(rootIndexes: FileDescription[], paths: string[])
...
+  const targetEthersVersion = require('../package.json').version
+
+  let major: number
+  let minor: number
+
+  try {
+    const { version } = require('typescript')
+    ;[major, minor] = version.split('.').map(Number)
+
+    if (major > requiredTSVersion.major || (major === requiredTSVersion.major && minor >= requiredTSVersion.minor)) {
```
krzkaczor, line 212: tbh i would probably extract this to a pure, isolated function and test only that one but this is fine too ;)

krzkaczor (approved):

PR #658, opened 2022-03-21
Write TypeChain just once

hasparus:
###  What's changed

At this moment our readme says:

> _TypeChain_
> _TypeChain_

After this change, there's only one _TypeChain_, the one in the image (and its alt text).


krzkaczor:
I did it for the SEO 😆 but probably having alt is good enough.

PR #665, opened 2022-03-24
Fix README typo

sirhashalot:


krzkaczor:
@sirhashalot thanks for catching this 👍

PR #674, opened 2022-04-15
Type migration in truffle-v5 example

bard:
Migration functions in the truffle-v5 example are cast, thus providing on type safety nor autocompletion. In the following, `deployer` has type `any`:

```ts
module.exports = function (deployer) {
   deployer.deploy(Migrations)
} as Truffle.Migration
```

This PR changes to a slightly longer format where `deployer`'s type is recognized as `Truffle.Deployer`:

```ts
const migration: Truffle.Migration = (deployer) => {
  deployer.deploy(Migrations);
};
module.exports = migration;
```


krzkaczor:
Thanks! 🙇

PR #681, opened 2022-04-27
Draft of Starknet target

maciejka:


packages/test-utils/package-lock.json
```diff
@@ -0,0 +1,191 @@
+{
```
krzkaczor, line 1: this lockfile shouldn't be needed.

.mocharc.js
```diff
@@ -5,5 +5,5 @@ module.exports = {
   extension: ['ts'],
   watchExtensions: ['ts'],
   spec: ['test/**/*.test.ts'],
-  timeout: 5000,
```
krzkaczor, line 8: can we override this setting only where it's needed? i guess in `target-starknet-test` package only

packages/target-starknet-test/test/InputReturnTypes.test.ts
```diff
@@ -0,0 +1,123 @@
...
+        contract.request_nested_structs({
+          p1: { x: 1, y: 2 },
+          p2: { x: 3, y: 4 },
+          extra: 5,
+        }),
+      ).not.toBeRejected()
+    })
+
+    it('Parsing the tuple in request', async () => {
+      await expect(contract.request_tuple([1, 2])).not.toBeRejected()
```
krzkaczor, line 57: i think we could drop `.not.toBeRejected ` everywhere here becuase if it would be rejected the whole test would bail 🤔

packages/target-starknet/src/StarknetTarget.ts
```diff
@@ -21,30 +23,48 @@ export class StarknetTarget extends TypeChainTarget {
...
+    const {
+      structTypes,
+      // constructorArgs,
+      functions,
+      imports,
+      impoort,
+    } = transformer(compiled.abi)
+
+    const ContractInterface = impoort('starknet', 'ContractInterface')
+    // const ContractFactory = impoort('starknet', 'ContractFactory')
```
krzkaczor, line 36: is this needed?

packages/target-starknet/src/StarknetTarget.ts
```diff
@@ -53,120 +73,159 @@ export class StarknetTarget extends TypeChainTarget {
...
+        const rets = returns(e, returnType)
+        return `${e.name}(${args}${args !== '' ? ', ' : ''}${opts}):${rets}`
       })
-      .join('\n')
   }
 
-  return {
-    requestImport,
-    generateImportStatements,
+  // function generateEventTypes(): string {
```
krzkaczor, line 115: should this be commented out?

krzkaczor (changes requested):

packages/target-starknet/src/StarknetTarget.ts
maciejka: You mean `ContractFactory`? it will be commented out when starknet.js start to interpret constructor arguments in the same way it interprets method args.

packages/target-starknet-test/test/InputReturnTypes.test.ts
maciejka: It gives us slightly more civilised error messages. Without the huge stacktrace in case test fails.

PR #682, opened 2022-05-01
[ethers-v5] Add initial support for custom errors

fredlacs:
This PR adds initial support for handling [custom errors](https://blog.soliditylang.org/2021/04/21/custom-errors/) with Typechain.
Still WIP but opening a draft PR to be able to get input from others on the approach (currently heavily inspired on https://github.com/dethcrypto/TypeChain/pull/457).

The idea here is to be able to handle errors in js/ts more cleanly when interacting with the chain. 
Although the functionality isn't here yet, it would be cool to be able to use these to have cleaner try catches when doing interactions with ethersjs

The current setup returns "undefined" for the error objects but is inspired in [this](https://github.com/ethers-io/ethers.js/blob/fc1e006575d59792fa97b4efb9ea2f8cca1944cf/packages/abi/src.ts/interface.ts#L260). A couple alternatives on how this could work (ie could rely on ethers and upstream something there or could just have a separate entrypoint). That said, I think only the typed error object exports themselves are already useful, could also have a utility function to get the signature / encode expected values


fredlacs:
I think what we'd want here is to cast the ErrorFragment to a typed version of the error

krzkaczor:
Related: https://github.com/dethcrypto/TypeChain/issues/667

krzkaczor (approved):
This looks reasonable. I am personally not a huge fan of custom errors and furthermore, ethersjs support is kinda minimal but the PR looks good!

Can you try fixing CI?

shashidhar771892:
CI test(14.x) the test was failed
Failing after 4m — test (14.x, ubuntu-latest

frontier159:
> This looks reasonable. I am personally not a huge fan of custom errors and furthermore, ethersjs support is kinda minimal but the PR looks good!

@krzkaczor out of interest - what's the reasoning for not being a fan of custom errors? It feels like they provide a better opportunity to handle errors in clients, and can also reduce bytecode in deployed contracts. But i'm interested in any downsides too before using in a project

shashidhar771892:
Error: Process completed with exit code 1.
We can add the 56s in the CI / Test (14.x ubuntu-latest)

wottpal:
This would be an awesome improvement for typechain and the web3 developer experience overall IMO. Really looking forward to see this one merged. Would also be interested in arguments against it, @krzkaczor.

GeraldHost:
Would love to see this merged, will unblock an issue with hardhat not having access to the errors in the interface provided by typechain and therefore not being able to use assertion helpers such as `revertedWithCustomError` in the chain helpers plugin they provide.

KholdStare:
@fredlacs @krzkaczor Is this pull request abandoned? Anything others can do to take it over the finish line?

gagiuntoli:
I am having problems understanding how custom-errors are handled. As far as I know, when a transaction fails, there is no information, e.g. on the tx receipt, about the error's reason. I have thought that the only way would be through transaction simulation and using directly a node with an EVM. How does ethers.js is going to handle these errors?

tfalencar:
+1 for this feature. Can someone please re-trigger CI? The logs for latest checks are too old and disappeared.

CJ42:
Having support for custom error in Typechain would be indeed very useful.
Would be great if someone could re-trigger the CI :)

KholdStare:
@fredlacs @krzkaczor Poking again, since it's been more than a year since this PR has been posted and issue https://github.com/dethcrypto/TypeChain/issues/667 has been raised.

emretepedev:
+1 for this feature.

Alivers:
Any updates?

Issue #692, opened 2022-05-15
Discriminate type fixes for typechian


ShravanSunder:
Config
- TypechainUserConfig is missing options

interface does not need contractName and shouldn't have it.
- ethers.interface is a representation of ABI standard and abstracts them
- contract and contractFactory are sufficient for discriminate union
- contracts contain ethers.interface as a prop

PR #693, opened 2022-05-15
#692/discriminate fixes

ShravanSunder:
- #692


ShravanSunder:
@krzkaczor 

Config
- TypechainUserConfig is missing options
- export default flag
interface does not need contractName and shouldn't have it.

- ethers.interface is a representation of ABI standard and abstracts them
- contract and contractFactory are sufficient for discriminate union
- contracts contain ethers.interface as a prop

ShravanSunder:
@krzkaczor anything else i need to do for this PR?

PR #700, opened 2022-05-24
Support promise inputs in ethers-v5 target

zemse:
Before:

```ts
const balanceOfOwner = await dai.balanceOf(await someContract.owner())
```

After: 

```ts
const balanceOfOwner = await dai.balanceOf(someContract.owner())
```


krzkaczor:
huh, i am somehow suprised that ethers.js supports promises instead of values :O We will need runtime tests for: a mocha test case where you would pass value and then pass promise of that value etc to test if this is really a case in runtime and to prevent regressions.

packages/target-ethers-v5-test/test/DataTypesInput.test.ts
```diff
@@ -79,6 +80,75 @@ describe('DataTypesInput', () => {
     typedAssert(await chain.contract.input_enum(1), 1)
   })
 
+  it('promise works', async function () {
+    const { contract } = chain
+
+    typedAssert(await chain.contract.input_uint8(getPromise('42')), 42)
```
zemse, line 86: @krzkaczor would typedAssert also execute the code here and compare runtime values? (I assumed that, if not the case then I'll add the tests)

zemse: btw yeah its a very underrated/underused thing in ethers that it resolves promise of values passed. Once I just passed a promise while in [REPL](https://github.com/zemse/ethers-repl), when I hit return I realised ahh this is going to fail, and the next second saw that it worked ✨

krzkaczor: is this documented somewhere?

krzkaczor: Yeah if @ricmoo says it's officially supported, then its fine. Thank you for clarifying that.

ricmoo: It is used quite extensively internal to ethers too.

I guess I only ever talked about it during presentations? I didn’t realize it was so secret. :)

PR #709, opened 2022-06-03
feat(codegen): reduce needed imports

sambacha:
Opening this PR to see if you are open to directly importing from @ethersproject packages rather than ethers umbrella.

This *may* be a breaking change? 

Here is a trivial example before we get too deep:
 
These two imports can be taken directly from the @ethersproject/contracts package. no need for umbrella import, should be a drop in.

next
```typescript
import type { BaseContract, Signer, utils } from "ethers";
```


sambacha:
Note. this a naive commit , do not merge

krzkaczor:
I think that the current way of importing stuff assumes less about the project that the user has set up. (what if the user uses only particular packages instead of the full ethers.js package?). For that reason, I think the current state of things is better. Thanks for the PR though!

PR #711, opened 2022-06-07
Feature/exclude gas from abi

frontier159:
Proposed fix for: https://github.com/dethcrypto/TypeChain/issues/677

Old vyper compiler versions generate `gas` estimates in the ABI. These members are:
- numbers. Since they're not strings they have incorrect types types when the _abi is passed to ethers `Contract` for example.
- The gas estimates are actually incorrect anyway - so even if they were stringified, `gasLimit` overrides would need to be set all over the place.

Easiest solution is to remove the `gas` field in the generated typechain factory code.

### DIFF
--- a/packages/target-ethers-v5/src/codegen/index.ts
+++ b/packages/target-ethers-v5/src/codegen/index.ts
@@ -240,6 +240,23 @@ export function codegenAbstractContractFactory(contract: Contract, abi: any): st
   `
 }
 
+// Earlier versions of the vyper compiler included the 'gas' estimates as a member
+// in the vyper-generated abi object.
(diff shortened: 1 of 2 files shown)


krzkaczor:
Thanks for working on this. My thoughts:
1) For sure we don't want to add ugly workarounds to generated code for all users. We should scope it only to parties impacted by this. Can we detect that `gas` field exists and only then remove it? Also  can we fix the abi build time instead of generating code to fix it during the runtime?
2) This needs a E2E test with harhdat-vyper.

I understand that you might not have a time to work on this but I can't merge it like this. We will get to this issue eventually.

PR #737, opened 2022-07-04
[ethers-v5] Separate import and type import

noy4:
This PR fixes:
1. importing types without 'import type' was problematic with svelte vite
2. importing not type object with 'import type' was problematic with svelte vite

I faced a same problem with this Issue: https://github.com/dethcrypto/TypeChain/issues/732

image

This problem for contract typings was fixed by this PR: https://github.com/dethcrypto/TypeChain/pull/636, but not for contract factory (maybe it was left out).

I simply turned everything in ethers to be imported with `import type` in `Some__factory.ts`, but faced another problem↓

image

So, I separated `import` and `import type` for ethers.

image

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -30,7 +30,7 @@ set the current solidity version.
 ### Local linking
 
 Run `pnpm build` to build all packages or `pnpm watch` to start watching. Then enter desired package directory and run
-`pmpm link`.
+`pnpm link`.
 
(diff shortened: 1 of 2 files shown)


tfalencar:
@krzkaczor do you see any issues with this PR?

vojtechsimetka:
Any ETA when this can be merged? I'll gladly help if there is some code needed.

PR #743, opened 2022-07-22
Use "import * as ___" with a separate "export { ___ }" declaration

### LINKED ISSUE #742: Use "import * as ___" with a separate "export { ___ }" declaration
In the first few lines of the generated `index.d.ts`, the folders are exported as a pair of `import` and `export`. For example:
``` typescript
import type * as chainlink from "./@chainlink";
export type { chainlink };
```
The exception is for the `factories` folder that is exported as:
``` typescript
export * as factories from "./factories";
```
It's also used to export other folders in TypeChain.

This causes the [api-exporter](https://api-extractor.com/) to error with the message `Error: The "export * as ___" syntax is not supported yet; as a workaround, use "import * as ___" with a separate "export { ___ }" declaration`.

Is it possible to always use the two-lined export for folders instead of the one-liner?

aalmada:
Use

```typescript
import type * as factories from "./factories";
export type { factories };
```

instead of

```typescript
export * as factories from "./factories";
```

A workaround because [api-exporter](https://api-extractor.com/) does not support the one-liner export.

Fixes #742 

Passed all the unit tests.


aalmada:
@krzkaczor Hi! Any chance there will be a new release soon containing this PR?

PR #761, opened 2022-09-03
[Hardhat plugin] Don't import libraries until they are needed

alcuadrado:
Hey @krzkaczor,

Just a quick PR optimizing this plugin. This makes projects using it run faster in general. You can appreciate the difference if you run a `npx hardhat --help` in a project using this version of the plugin and `latest`.

The difference is not huge, but it compounds with other plugins, so I'm doing this for every plugin included in the hardhat toolbox.

Also, I don't think lodash is needed anymore, so I removed it.

### DIFF
--- a/.changeset/light-hairs-chew.md
+++ b/.changeset/light-hairs-chew.md
@@ -1,0 +1,5 @@
+---
+"@typechain/hardhat": patch
+---
+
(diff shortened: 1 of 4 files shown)


krzkaczor:
Cool stuff, thank you! I am gonna release it in a sec.

PR #764, opened 2022-09-07
Set signerOrProvider optional

Rubilmax:
Hi,

Here's the Contract constructor signature from ethers v5:
```typescript
constructor(addressOrName: string, contractInterface: ContractInterface, signerOrProvider?: Signer | Provider)
```

I was thus wondering why TypeChain forces `signerOrProvider` to be defined, whereas `ethers` does not (and we've got a use-case where we don't want to define a signer nor a provider)

So here's a quick PR to fix this :)

### DIFF
--- a/packages/hardhat-test/typechain-types/factories/Counter__factory.ts
+++ b/packages/hardhat-test/typechain-types/factories/Counter__factory.ts
@@ -105,7 +105,7 @@ export class Counter__factory extends ContractFactory {
   }
   static connect(
     address: string,
-    signerOrProvider: Signer | Provider
+    signerOrProvider?: Signer | Provider
   ): Counter {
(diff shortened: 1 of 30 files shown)


PaulFrambot (approved):

MerlinEgalite (approved):
LGTM 🔥

julien-devatom (approved):

mihaic195 (approved):

mihaic195:
Thanks for this! Much needed!

iainnash:
+1 to this

mihaic195:
This has been open for almost 2 months and the problem is still there.
@krzkaczor can we get this merged, please?

Rubilmax:
Up

Issue #772, opened 2022-09-15
Typechain fails where there are no contracts in a shared parent directory


alexdupre:
Here is a sample artifacts hierarchy structure to reproduce the bug:

```
contracts/1/interfaces/IOne.sol/IOne.json
contracts/2/interfaces/ITwo.sol/ITwo.json
```

Running typechain on a single directory works fine, `typechain --target ethers-v5 --out-dir './typechain' './contracts/1/**/*.json'` generates the following files:
```
./typechain/common.ts
./typechain/factories/index.ts
./typechain/factories/IOne__factory.ts
./typechain/index.ts
./typechain/IOne.ts
```

On the other hand, trying to generate the types for both contracts concurrently with `typechain --target ethers-v5 --out-dir './typechain' './contracts/**/*.json'` generates the following error:
```
Error occured:  'from' expected. (4:24)
  2 | /* tslint:disable */
  3 | /* eslint-disable */
> 4 | import type * as  from './..';
    |                        ^
  5 | export type {  };
  6 | export * as 1 from './1';
  7 | export * as 2 from './2';
```

and some files are generated outside the requested `out-dir`:

```
./1/interfaces/IOne.ts
./2/interfaces/ITwo.ts
./typechain/common.ts
```

alexdupre:
The issue seems to be in the `lowestCommonPath` function that returns the wrong directory.

In particular this line is wrong: https://github.com/dethcrypto/TypeChain/blob/master/packages/typechain/src/utils/files/lowestCommonPath.ts#L3

because in the above scenario it returns `contracts/interfaces` as the lowest common path (a directory that doesn't even exist) instead of `contracts`.

ondratra:
I encountered this bug when working with HardHat, and it took me a while to realize the problem was not my HardHat settings but actually in TypeChain library. Thx for the fix @alexdupre!

I can confirm the related PR #773 fixes the issue. Can we get it merged, please? (@krzkaczor, sry for pinging, but can you do that, please?)

kazazor:
Same here, this pR #773 fixes it

krzkaczor:
@kazazor, thanks for mentioning this. I am merging the fix now.

ondratra:
I can confirm the issue has been solved in https://github.com/dethcrypto/TypeChain/pull/858 and it can be closed. @krzkaczor if you will, please :-)

PR #773, opened 2022-09-19
Fix the `lowestCommonPath` function, by stopping at first different sub-path

alexdupre:
This fixes issue #772


krzkaczor:
Merged as https://github.com/dethcrypto/TypeChain/pull/858

PR #779, opened 2022-10-06
Add struct parsing to @typechain/web3-v1

mihai9-lab:
This PR adds support for struct parsing to @typechain/web3-v1, similarly to @typechain/ethers-v5.

Target from this PR is on [NPM ](https://www.npmjs.com/package/typechain-target-web3-v1-3mihai3) if you want to try it out

### DIFF
--- a/.changeset/pink-ads-rhyme.md
+++ b/.changeset/pink-ads-rhyme.md
@@ -1,0 +1,5 @@
+---
+'@typechain/web3-v1': minor
+---
+
+Add struct parsing and typing for web3 library
--- a/packages/target-web3-v1-test/types/v0.6.4/DataTypesInput.ts
+++ b/packages/target-web3-v1-test/types/v0.6.4/DataTypesInput.ts
@@ -21,6 +21,46 @@ export interface EventOptions {
   topics?: string[];
 }
 
+export declare namespace StructsLib1 {
+  export type InfoStruct =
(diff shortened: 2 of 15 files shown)


sigmachirality:
+1 to this fix/issue, we ran into TS types missing Solidity struct key names in trying to integrate types for this target into our webapp. I hope some form of this get merged soon!

sigmachirality:
Hi @krzkaczor, can you provide some feedback?

We use web3-v1 in prod and we often access the struct objects returned from function calls, so having the properties typed by name is imperative before we can adopt this typechain target into our workflow.

Alternatively, would you prefer an implementation which generates types as 
```
export type ReturnStruct = GenericReturnStruct<{
  property1: string;
  property2: string;
  0: string;
  1: string;
}>;
```
Similar to how Events types are generated in this target, as opposed to `ReturnStructAsAnObject & ReturnStructAsAnArray` (how they are generated in the ethers target)? If so, I'm down to work on this PR on more to get it into a state that's able to be merged.

Please let me know, and thanks in advance!

mihai9-lab:
Hi @sigmachirality, i think that the issues you mentioned are already covered by this PR. If not, could you elaborate a bit more? 

A few days ago i published this as a package on npm, under the name `typechain-target-web3-v1-3mihai3`, you could try adding it to your project and seeing if it works for you. Just set the target to `web3-v1-3mihai3`.

PR #783, opened 2022-10-14
ethers: add const assertion for exported ABI

kdembler:
Fixes https://github.com/dethcrypto/TypeChain/issues/782


ziad-saab:
@krzkaczor what would it take to get this merged?

kdembler:
@krzkaczor Could you please take a look? Without this change, projects depending on Typechain bindings will not be able to use `wagmi` 0.7+ in a type-safe way.

dennzimm:
+1

royalaid:
I would also like to echo support for this, it seems to be the easiest way to achieve a nice bridge between wagmi and typechain.

krzkaczor (approved):

krzkaczor:
Thanks for this @kdembler. It's released as part of:
https://github.com/dethcrypto/TypeChain/releases/tag/%40typechain%2Fhardhat%406.1.5

Sorry that it took a while.

PR #784, opened 2022-10-15
[web3.js] support BigNumber.js

DanielZlotin:
[BigNumber.js](https://github.com/MikeMcl/bignumber.js/) is a small and very popular library when developing for web3.
`web3.js` supports BigNumber as well as BN as input values to methods, events etc; so this PR adds support for the types generated by TypeChain to accept either.


paulmillr:
-1

all such libraries are being replaced by native bigints

PR #787, opened 2022-10-29
chore: fixed name of starknet.js target

ivpavici:
Name of the target was @typechain/starknet.js but should be @typechain/starknet


ivpavici:
Same PR as https://github.com/dethcrypto/TypeChain/pull/748

PR #799, opened 2023-01-07
Add node16-modules flag to support Node16/NodeNext modules resolutions

silasdavis:
TypeScript's Node16/NodeNext module resolution brings uniform native ESM module support to Node/ESM environments as well as support for multiple entrypoints via package.json `exports`. However it explicitly requires relative imports to use file extensions for modules, usually '.js'.

This change introduces a `node16Modules` codegen config. Currently it just adds the module suffix `.js` to imports and exports where needed.

Without this it is impossible to use Typechain generated code with tsconfig moduleResolution: Node16 or NodeNext.

Signed-off-by: Silas Davis <silas.davis@monaxlabs.io>

### DIFF
--- a/packages/target-ethers-v5/src/codegen/index.ts
+++ b/packages/target-ethers-v5/src/codegen/index.ts
@@ -116,9 +116,10 @@ export function codegenContractTypings(contract: Contract, codegenConfig: Codege
     };
   }`
 
-  const commonPath = contract.path.length
-    ? `${new Array(contract.path.length).fill('..').join('/')}/common`
-    : './common'
+  const moduleSuffix = codegenConfig.node16Modules ? '.js' : ''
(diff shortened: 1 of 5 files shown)


krzkaczor:
Thanks @silasdavis! Can u take a look at failing tests?

silasdavis:
Oops, yeah I will try to get round to that

pablomendezroyo:
kind bump for this feature. The current documentation is encouraging users that using @typechain/ethers-v6 set the `moduleResolution` to either: `nodenext` or `node16`. This is incompatible with the types generated by typechain due to the missing extension `.js`

krzkaczor:
Merged as: https://github.com/dethcrypto/TypeChain/pull/840

PR #801, opened 2023-01-23
Add struct parsing to @typechain/web3-v1 (#779)

krzkaczor:


krzkaczor:
@mihai9-lab  there are still some problems with tests. Can u take a look?

packages/target-web3-v1/src/codegen/types.ts
```diff
@@ -57,18 +66,87 @@ export function codegenOutputType(evmType: EvmOutputType): string {
...
 }
 
-export function codegenTupleType(tuple: TupleType, generator: (evmType: EvmType) => string) {
-  return '[' + tuple.components.map((component) => generator(component.type)).join(', ') + ']'
+export function codegenArrayOrTupleType(item: string, length?: number) {
+  if (length !== undefined && length < 6) {
+    return `[${Array(length).fill(item).join(', ')}]`
+  } else {
+    return `${item}[]`
+  }
```
sigmachirality, line 89: Hi @mihai9-lab, we ran across a bug on this line specifically. In the case where item is a string declaring a union type (for example: `string | number | BN`), just appending `[]` to this type results in `string | number | BN[]` instead of the intended type `(string | number | BN)[]` or perhaps `string [] | number[] | BN[]` (I prefer the first solution, as it allows mixing of types in input arrays).

I was able to fix this with the following code:
```
if (item.contains("|")) {
  item = `(${item})`
}
```

However, it may be cleaner to parametrize this function with another boolean and determine whether item is a union upstream of this function, and/or accept length and this proposed boolean as properties of an options object.

packages/target-web3-v1-test/types/v0.6.4/Events.ts
```diff
@@ -45,8 +56,8 @@ export type Event3_uint256 = ContractEventLog<{
   0: string;
 }>;
 export type Event4 = ContractEventLog<{
-  data: [string, string];
-  0: [string, string];
+  data: Events.EventDataStructOutput;
+  0: Events.EventDataStructOutput;
```
sigmachirality, line 60: nit: I prefer keeping these types flat for ease of use during development (would prefer to see that the type is a [string, string] instead of a Events.EventDataStructOutput when hovering over it, for example)

packages/target-web3-v1/src/codegen/events.ts
```diff
@@ -73,7 +73,7 @@ function codegenEventDeclaration(event: EventDeclaration, overloadedName?: strin
 
 function codegenOutputTypesForEvents(outputs: EventArgDeclaration[]): string {
   return `{
-    ${outputs.map((t) => t.name && `${t.name}: ${codegenOutputType(t.type)}, `).join('')}
-    ${outputs.map((t, i) => `${i}: ${codegenOutputType(t.type)}`).join(', ')}
+    ${outputs.map((t) => t.name && `${t.name}: ${codegenOutputType({ useStructs: true }, t.type)}, `).join('')}
+    ${outputs.map((t, i) => `${i}: ${codegenOutputType({ useStructs: true }, t.type)}`).join(', ')}
```
sigmachirality, line 77: any chance we can reverse this function's type params to `codegenOutputType(type, options)` instead of `codegenOutputType(options, type)`? The first is more likely to not break code, since it doesn't change the position of the type param.

sigmachirality: Seems like most codegen functions that take in an options param also take it as the last parameter

packages/target-web3-v1/src/codegen/types.ts
```diff
@@ -1,26 +1,32 @@
...
 
-export function codegenInputTypes(input: AbiParameter[]): string {
+import { STRUCT_INPUT_POSTFIX, STRUCT_OUTPUT_POSTFIX } from '../common'
+
+interface GenerateTypeOptions {
+  returnResultObject?: boolean
+  useStructs?: boolean // uses struct type for first depth, if false then generates first depth tuple types
+}
+
+export function codegenInputTypes(options: GenerateTypeOptions, input: Array<AbiParameter>): string {
```
sigmachirality, line 11: Same feedback for codegenInputTypes (prefer `codegenInputTypes(input, options?)`)

sigmachirality (changes requested):
Excited to see this being worked on! Commenting a few nits/flagging a bug we ran into trying to use @mihai9-lab's target

packages/target-web3-v1/src/codegen/types.ts
sigmachirality: Seems like the original code here just wrapped all types in parenthesis regardless of whether the type was simple or a union type. I leave the exact form of the solution up to you!

sigmachirality:
@mihai9-lab when do you think you'll have bandwidth to work on this change? Want to get this through pretty urgently, can I work on this now if you don't have time soon?

mihai9-lab:
> @mihai9-lab when do you think you'll have bandwidth to work on this change? Want to get this through pretty urgently, can I work on this now if you don't have time soon?

Hey, sorry for not replying sooner. Currently I'm pretty busy and won't be able to work on this for at least a week or two. 
I have no issues with you working on this, so feel free to do so

sigmachirality:
The approach of this PR might be a bit flawed: as seen here: https://web3js.readthedocs.io/en/v1.8.1/web3-eth-contract.html#methods-mymethod-call

"Promise returns Mixed: The return value(s) of the smart contract method. If it returns a single value, it’s returned as is. If it has multiple return values they are returned as an object with properties and indices"

Since the return is "an object with properties and indices", we cannot return the type of a struct as a union of an array and an object, as this will annotate the object with all the members of an array (for example, .concat), which do not actually exist on the object.

mihai9-lab:
Yeah i didn't expect it to behave like that, it seems that typescript does assume object has all properties of an array(although it isn't one). 

I see 2 possible approaches to deal with this - one being to change array generation to object generation, other being to leave it as is but add an additional "converter" type which would exclude unwanted array properties.

I made a quick array to objected converter type as a proof of concept if going with second option, here's a [playground](https://www.typescriptlang.org/play?#code/C4TwDgpgBAggTnAhiAKgewPICMBWEDGwAPClBAB7AQB2AJgM5SLUgDaAugHxQC8UA3gCgoUVgGkoAS2pQAouXwAbAK60IRANYQQaAGZQUAGi079zNl3YAuFOPaCAvoMGhIUAMrA48OL1H0vaQBzQwC4YNDA6iD7F3BoT28EAGE0agA3CDgqWj8fZHRsPEIiRJ9OZ1doeUQAWzBFCCK-IREKAEYrMIjhMnIAJitWbujI8Oj7JyqoVNqsaQhaHz9hqJCa+saiwzKEWOnZ+epFn1SMrJy8hALMXAJiQ4WlhArBAHo3qFrkLGhgAAtJIx8MwoL8oMp6BBdMpFExGAB3CCKRQAfjibnyqFuxWAACUCMo4PRJJkSH0qHRGOYONw+K1RBJpHIFCo1JptHoDMZOWYWLTrLYxOwKTQGEx+exUVjCncSkKuDY7I5KvEZmg5k9TmlMtlFgT8ESSZkrkhsUV7gajaT1I9js84JwgA) for that. Maybe this is a bit hacky, so i'd like to see what you and the community thinks is the best approach.

PR #811, opened 2023-02-24
fix(ethers-v6): TypedContractEvent should extend EventLog

Songkeys:
This is to align with v5:

```
import type { Event } from "ethers";

export interface TypedEvent<
  TArgsArray extends Array<any> = any,
  TArgsObject = any
> extends Event {
  args: TArgsArray & TArgsObject;
}
```

so that we can type the event log with it.

### DIFF
--- a/packages/target-ethers-v6/static/common.ts
+++ b/packages/target-ethers-v6/static/common.ts
@@ -11,7 +11,7 @@ export interface TypedContractEvent<
   InputTuple extends Array<any> = any,
   OutputTuple extends Array<any> = any,
   OutputObject = any,
-> {
+> extends EventLog {
   (...args: Partial<InputTuple>): TypedDeferredTopicFilter<TypedContractEvent<InputTuple, OutputTuple, OutputObject>>
   name: string
   fragment: EventFragment


zemse:
There is a `TypedEventLog` in the same file, which is the return type for queryFilter. The `TypedEvent` in v5 is similar to `TypedEventLog` in v6.

However, `TypedContractEvent` is a different thing, it is a proxy function that can be used to generate a filter. Maybe we can change its name if it's not the best.

Songkeys:
How do I type the event log?

In v5, I type them like this:

```ts
    import type { MintEvent } from '../Abi'


    const targetTopicHash = keccak256(toUtf8Bytes('Mint(uint256,uint256,uint256)')

    // txReceipt.logs
    const _logs = logs.filter((log) => log.topics[0] === targetTopicHash)

    if (_logs.length === 0) {
      throw new Error(`Log with topic ${filterTopic} not found`)
    }

    const log = _logs[0]

    return this.contract.interface.parseLog(log) as unknown as MintEvent
```

In v6, the `MintEvent` (`XxxEvent`) turns into an namespace which includes an `Event` type:

```ts
export namespace MintEvent {
  export type InputTuple = [
    to: BigNumberish,
    tokenId: BigNumberish,
    tokenNumber: BigNumberish
  ];
  export type OutputTuple = [to: bigint, tokenId: bigint, tokenNumber: bigint];
  export interface OutputObject {
    to: bigint;
    tokenId: bigint;
    tokenNumber: bigint;
  }
  export type Event = TypedContractEvent<InputTuple, OutputTuple, OutputObject>;
  export type Filter = TypedDeferredTopicFilter<Event>;
}
```

But the `MintEvent.Event` type (`TypedContractEvent`) doesn't extends `EventLog` to be compatible with `LogDescription`.

I tried using `TypedEventLog<MintEvent.Event>` but got this error:

```
type MintEvent.Event = TypedContractEvent<MintEvent.InputTuple, MintEvent.OutputTuple, MintEvent.OutputObject>
Type 'Event' does not satisfy the constraint 'TypedContractEvent<any, any, any>'.
  Types of parameters 'to' and 'args' are incompatible.
    Type 'Partial<any>' is not assignable to type '[to?: BigNumberish | undefined, tokenId?: BigNumberish | undefined, tokenNumber?: BigNumberish | undefined]'.ts(2344)
```

`TypedEventLog<TransferEvent.Event>` works fine actually.

Do you have any suggestions? thanks.

zemse:
> But the MintEvent.Event type (TypedContractEvent) doesn't extends EventLog to be compatible with LogDescription.

Yea, `MintEvent.Event` will not work but `TypedEventLog<MintEvent.Event>` should work since it extends EventLog. To prevent further confusion I think we should rename this to something else.

> I tried using TypedEventLog<MintEvent.Event> but got this error
> TypedEventLog<TransferEvent.Event> works fine actually.

Do you mean Transfer event works but Mint event gives type errors?

Can you try this once?

```ts
const parsedLog = parseLog(receipt.logs, 'Mint(uint256,uint256,uint256)') as unknown as TypedEventLog<MintEvent.Event>[]
```

You should be able to access `.args` since it is available in both EventLog and LogDescription. But for full compatibility (without requiring to cast into unknown), I can add a TypedLogDescription here.

Songkeys:
> Do you mean Transfer event works but Mint event gives type errors?

Yes. I pasted the error in the above comment. I paste here again:

The `MintEvent` generated for me is:

```ts
export namespace MintEvent {
  export type InputTuple = [
    to: BigNumberish,
    tokenId: BigNumberish,
    tokenNumber: BigNumberish
  ];
  export type OutputTuple = [to: bigint, tokenId: bigint, tokenNumber: bigint];
  export interface OutputObject {
    to: bigint;
    tokenId: bigint;
    tokenNumber: bigint;
  }
  export type Event = TypedContractEvent<InputTuple, OutputTuple, OutputObject>;
  export type Filter = TypedDeferredTopicFilter<Event>;
}
```

The error is:

```
type MintEvent.Event = TypedContractEvent<MintEvent.InputTuple, MintEvent.OutputTuple, MintEvent.OutputObject>
Type 'Event' does not satisfy the constraint 'TypedContractEvent<any, any, any>'.
  Types of parameters 'to' and 'args' are incompatible.
    Type 'Partial<any>' is not assignable to type '[to?: BigNumberish | undefined, tokenId?: BigNumberish | undefined, tokenNumber?: BigNumberish | undefined]'.ts(2344)
```

PR #823, opened 2023-04-12
Remove obsolete peer dependency `@ethersproject/bytes@^5.0.0` from `target-ethers-v5`

### LINKED ISSUE #822: Unmet peer dependency `@ethersproject/bytes@^5.0.0`
If using it via e.g. Hardhat (`npm install --save-dev typechain @typechain/hardhat @typechain/ethers-v5`) you won't need the peer dependency `@ethersproject/bytes` and the `yarn` installation will issue the following warning:

```bash
@typechain/ethers-v5@10.2.0" has unmet peer dependency "@ethersproject/bytes@^5.0.0"
```

> You can reproduce this behavior e.g. [here](https://github.com/pcaversaccio/hardhat-project-template-ts/actions/runs/4667684060/jobs/8263791566#step:7:11).

My question is whether we could get completely rid of this peer dependency or if there are some specific use cases where this is required.

pcaversaccio:
Fixes #822. No changeset added since it will not cause an immediate version bump.

### DIFF
--- a/.changeset/funny-cobras-invent.md
+++ b/.changeset/funny-cobras-invent.md
@@ -1,0 +1,5 @@
+---
+"@typechain/ethers-v5": patch
+---
+
+Remove obsolete peer dependency `@ethersproject/bytes@^5.0.0` from `target-ethers-v5`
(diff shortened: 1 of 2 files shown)


pcaversaccio:
@krzkaczor let me know if a changeset is required here.

pcaversaccio:
@zemse can we get this merged?

krzkaczor:
@pcaversaccio sorry that it took so long. Merged!

PR #832, opened 2023-05-10
fix type ContractRunner

### LINKED ISSUE #821: ContractRunner should be imported using `import type`
The issue is that my Svelte is not compiling due to `ContractRunner` in the factory does not get imported using `import type`. I think this is related to #636, but the fix did not include ethers-v6 target. Here is the code sample generated for ethers-v6 target:

https://github.com/dethcrypto/TypeChain/blob/2b29791881cfb87bfff62b31f375943c782a9905/packages/target-ethers-v6-test/types/factories/v0.8.9/ISimpleToken__factory.ts

The generated code should add `type` to `ContractRunner`

```diff
-import { Contract, Interface, ContractRunner } from "ethers";
+import { Contract, Interface, type ContractRunner } from "ethers";
```

clouds56:
fixes #821

### DIFF
--- a/.changeset/ethers-v6-runner.md
+++ b/.changeset/ethers-v6-runner.md
@@ -1,0 +1,5 @@
+---
+"@typechain/ethers-v6": patch
+---
+
+fix import type ContractRunner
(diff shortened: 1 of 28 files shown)


krzkaczor:
I needed to revert this PR because CI still fails. I will take a closer look tmrw.

krzkaczor:
Thanks! Merged again as: https://github.com/dethcrypto/TypeChain/pull/837

clouds56:
I wonder what happens internally.

PR #840, opened 2023-05-24
Add node16-modules flag to support Node16/NodeNext modules resolutions

krzkaczor:
Supersedes: https://github.com/dethcrypto/TypeChain/pull/799


mouchegmouradian:
The same thing needs to be done for ethers-v6

chmanie:
> The same thing needs to be done for ethers-v6

On it

PR #843, opened 2023-05-30
Add support for ethers v6 in @typechain/hardhat

### LINKED ISSUE #842: Add support for hardhat + ethers v6
I'm working on a new version of the `hardhat-ethers` plugin and I want to make sure that it works well with typechain. From what I can see, typechain's hardhat plugin hasn't been updated yet (which makes sense, because we haven't updated our plugin :sweat_smile:).

Some comments from looking at the codebase:
- The version of ethers used in `@typechain/hardhat` should be updated, and the dependency on `@typechain/ethers-v5` should be replaced with `@typechain/ethers-v6` 
- I think some peer dependencies (`@ethersproject/abi` and `@ethersproject/providers`) in `@typechain/hardhat` could be dependencies. This is no strictly necessary, but reducing peer dependencies prevents conflicts down the road.
- `codegen/hardhat.ts` in `@typechain/target-ethers-v6` will need to be updated. Here I think we'll have to coordinate, but at the very least:
  - The new version of the plugin will be called `@nomicfoundation/hardhat-ethers`, not `@nomiclabs`.
  - `getContractAt` will also accept an `Addressable` as the address, so those types should be changed to `string | Addressable`.

I'll be working on this today and I'll comment here anything else I find.

fvictorio:
Closes #842.

Tentative changes. I tried this locally and seems to work.

Something I don't like about my implementation is that the contract-specific types for `deployContract` don't have the right constructor parameters types. I'm not sure how to do this properly, but it seems to me that `generateHardhatHelper` should receive more information about the contracts to be able to do that, and not just the names.

I also changed some docs to use our chai matchers instead of waffle, since that's the approach we recommend, and because the chai matchers will have support for ethers v6 (and waffle doesn't yet afaik).


examples/hardhat-truffe-v5/package.json
```diff
@@ -14,7 +14,7 @@
     "@nomiclabs/hardhat-etherscan": "^2.1.1",
     "@nomiclabs/hardhat-truffle5": "^2.0.0",
     "@nomiclabs/hardhat-web3": "^2.0.0",
-    "@typechain/hardhat": "workspace:^7.0.0",
+    "@typechain/hardhat": "workspace:^8.0.0",
```
krzkaczor, line 17: why change this?

packages/hardhat/test/fixture-projects/hardhat-project/typechain-types/common.ts
```diff
@@ -1,33 +1,69 @@
...
-import type { Listener } from "@ethersproject/providers";
-import type { Event, EventFilter } from "ethers";
-
-export interface TypedEvent<
-  TArgsArray extends Array<any> = any,
-  TArgsObject = any
-> extends Event {
-  args: TArgsArray & TArgsObject;
-}
+/* Autogenerated file. Do not edit manually. */
```
krzkaczor, line 4: I think thit is a slight bug in ethers-v6 target that causes double preamble. Wanna have a look?

examples/hardhat-truffe-v5/package.json
fvictorio: My bad, this is a leftover of when I was trying things locally with verdaccio. I'll undo it.

krzkaczor: @fvictorio note i fixed this and all other issues in: https://github.com/dethcrypto/TypeChain/pull/845

packages/hardhat/test/fixture-projects/hardhat-project/typechain-types/common.ts
fvictorio: Done.

PR #845, opened 2023-06-03
Add support for ethers v6 in @typechain/hardhat #843

krzkaczor:
@fvictorio

Supersedes https://github.com/dethcrypto/TypeChain/pull/843


examples/hardhat/test/counter.ts
```diff
@@ -20,12 +20,12 @@ describe('Counter', () => {
     const counterFactory = await ethers.getContractFactory('Counter', signers[0])
 
     counter = await counterFactory.deploy()
-    await counter.deployed()
+    await counter.waitForDeployment()
```
krzkaczor, line 23: @fvictorio  have a look at changes in this file

packages/hardhat-test/test/hardhat.test.ts
```diff
@@ -21,7 +21,7 @@ describe('TypeChain x Hardhat', () => {
 
     const counterFactory1 = new Counter__factory()
     const counterFactory2 = new Counter__factory(deployer)
-    const counterFactory3 = new Counter__factory(artifact.abi, artifact.bytecode)
-    const counterFactory4 = new Counter__factory(artifact.abi, artifact.bytecode, deployer)
+    // const counterFactory3 = new Counter__factory(artifact.abi, artifact.bytecode)
```
krzkaczor, line 24: @fvictorio  have a look at changes in this file

examples/hardhat/test/counter.ts
fvictorio: LGTM, although I'd rather do just this:

```ts
    counter = await ethers.deployContract('Counter')
```

I'd also remove the `signers[0]` part because the first signer is used by default (unless you are being explicit on purpose).

packages/hardhat-test/test/hardhat.test.ts
fvictorio: I think you can uncomment these two lines and they will compile if you change this import:

```ts
import type { Interface } from '@ethersproject/abi'
```

to this:

```ts
import type { Interface } from 'ethers'
```

krzkaczor: Magic

examples/hardhat/test/counter.ts
krzkaczor: Good call. It's much simpler now. I want to keep these examples as a sort of "preferred usage" guide.

PR #848, opened 2023-06-21
Add directory exports module suffix

### LINKED ISSUE #819: [ERR_UNSUPPORTED_DIR_IMPORT] ES Module factory exports
**Note**: I am uncertain if this is an issue with`typechain` or `@typechain/ethers-v5`. This issue might also exist with `@typechain/web3-v1`
Recently upgraded typechain in my project to the versions below and I appreciate the modularized exports as we are looking to export typechain types in our contract's npm package. However there is a slight issue when using ESModules due to incorrect exports that do not explicit use the file extensions. 
Used versions:
```json
{
"typechain": "8.1.1",
"@typechain/ethers-v5": "10.2.0",
"@typechain/web3-v1": "6.0.2",
}
```

Below a short example snippet of the typechain generated `index.ts` when using openzeppelin.
```typescript
//index.ts
import type * as openzeppelin from "./@openzeppelin";
export type { openzeppelin };
export * as factories from "./factories";
export type { IERC1155Upgradeable } from "./@openzeppelin/contracts-upgradeable/token/ERC1155/IERC1155Upgradeable";
export { IERC1155Upgradeable__factory } from "./factories/@openzeppelin/contracts-upgradeable/token/ERC1155/IERC1155Upgradeable__factory";
```
We then export these types in our package so developers can easily use them.
```typescript
export * as TypechainEthers from "./typechain/ethers/index.js"
```

There is a slight issue when using ESModules. When using ESModules, js files **MUST** be imported with explicit file name extensions. However this only applies to importing variables and **NOT** types in TS files (as these have no runtime value). Therefore in the snippet above, while the contract types are properly exported, the factory exports (both `factories` and  `IERC1155Upgradeable__factory`) don't just export types but rather a class that mimics ethers factory but with proper types. Therefore the import breaks because the factory is not exported with explicit reference to the `.js` file. Here would be an example fix to the above snippet.

```typescript
//index.ts
import type * as openzeppelin from "./@openzeppelin";
export type { openzeppelin };
export * as factories from "./factories/index.js"; //add "/index.js"
export type { IERC1155Upgradeable } from "./@openzeppelin/contracts-upgradeable/token/ERC1155/IERC1155Upgradeable";
export { IERC1155Upgradeable__factory } from "./factories/@openzeppelin/contracts-upgradeable/token/ERC1155/IERC1155Upgradeable__factory.js"; //add ".js"
```

Please let me know if you have an idea where the logic for the export file generation is and I would be open to looking into making the patch.

chmanie:
This adds a proper suffix to directory imports as well (as long as they are actual files and not types).

This should close #819.

### DIFF
--- a/.changeset/lazy-eggs-brush.md
+++ b/.changeset/lazy-eggs-brush.md
@@ -1,0 +1,6 @@
+---
+'typechain': patch
+---
+
+Add the node16 moduleSuffix also to directories (including the `index` filename) as in ESM directory root files have to
+be explicitly stated (including their file extension).
--- a/examples/ethers-v5-nodenext/types/ethers-contracts/index.ts
+++ b/examples/ethers-v5-nodenext/types/ethers-contracts/index.ts
@@ -2,5 +2,5 @@
 /* tslint:disable */
 /* eslint-disable */
 export type { Dai } from "./Dai.js";
-export * as factories from "./factories";
+export * as factories from "./factories/index.js";
 export { Dai__factory } from "./factories/Dai__factory.js";
--- a/packages/typechain/src/codegen/createBarrelFiles.ts
+++ b/packages/typechain/src/codegen/createBarrelFiles.ts
@@ -59,6 +59,9 @@ export function createBarrelFiles(
             `export type { ${namespaceIdentifier} };`,
           ].join('\n')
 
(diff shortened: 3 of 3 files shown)


krzkaczor:
Thanks! It will be released is a second.

PR #852, opened 2023-07-16
fix: [ethers-v6] codegen/hardhat.ts DeployContractOptions expected

DainisGorbunovs:
## Issue
[@nomicfoundation/hardhat-ethers@3.0.1](https://github.com/NomicFoundation/hardhat/releases/tag/%40nomicfoundation%2Fhardhat-ethers%403.0.1) added support for transaction overrides in the `deployContract` helper.

Specifically, these are the changes  https://github.com/NomicFoundation/hardhat/commit/d8056fb8bc1116338f61974a8a44cd0defcb2555.

Here's issue raised in `hardhat` repository:
- https://github.com/NomicFoundation/hardhat/issues/4118

## Fix
Replaced `FactoryOptions` with `DeployContractOptions`, and added import for it. Reran the tests to fix other files.

## Context
A new project is created:
```sh
npm init -y
npm install --save-dev hardhat
npx hardhat # Create a TypeScript project
npx hardhat compile
```

It installs `@nomicfoundation/hardhat-ethers" v3.0.4. As hardhat uses [@typechain/ethers-v6](https://github.com/NomicFoundation/hardhat/blob/c2943366d4f16282c10f92f9f216ac08130a0eb4/yarn.lock#L1672), I've only updated `packages/target-ethers-v6/src/codegen/hardhat.ts`, which should be enough to resolve the issue.

The provided `scripts/deploy.ts` contains:
```ts
import { ethers } from "hardhat";

async function main() {
  const currentTimestampInSeconds = Math.round(Date.now() / 1000);
  const unlockTime = currentTimestampInSeconds + 60;

  const lockedAmount = ethers.parseEther("0.001");

  const lock = await ethers.deployContract("Lock", [unlockTime], {
    value: lockedAmount,
  });
```

IDE complains about `{ value: lockedAmount }`. Third parameter in `ethers.deployContract` used to be for `signerOrOptions?: ethers.Signer | FactoryOptions` but was changed to `signerOrOptions?: ethers.Signer | DeployContractOptions`

Visual Studio Code:
```
No overload matches this call.
  Overload 1 of 4, '(name: "Lock", args: any[], signerOrOptions?: Signer | FactoryOptions | undefined): Promise<Lock>', gave the following error.
    Argument of type '{ value: bigint; }' is not assignable to parameter of type 'Signer | FactoryOptions | undefined'.
      Object literal may only specify known properties, and 'value' does not exist in type 'Signer | FactoryOptions'.
```

### DIFF
--- a/examples/hardhat/package.json
+++ b/examples/hardhat/package.json
@@ -11,7 +11,7 @@
     "typecheck": "pnpm compile && pnpm test && pnpm tsc --noEmit"
   },
   "devDependencies": {
-    "@nomicfoundation/hardhat-ethers": "^3.0.0",
+    "@nomicfoundation/hardhat-ethers": "^3.0.4",
     "@nomicfoundation/hardhat-chai-matchers": "^2.0.0",
     "@typechain/ethers-v6": "workspace:^0.4.0",
     "@typechain/hardhat": "workspace:^8.0.0",
--- a/examples/hardhat/test/counter.ts
+++ b/examples/hardhat/test/counter.ts
@@ -2,6 +2,7 @@ import chai from 'chai'
 import chaiAsPromised from 'chai-as-promised'
 import { solidity } from 'ethereum-waffle'
 import { ethers } from 'hardhat'
+import { DeployContractOptions } from "@nomicfoundation/hardhat-ethers/types"
 
 import type { Counter } from '../typechain-types'
 
@@ -17,13 +18,15 @@ describe('Counter', () => {
     const signers = await ethers.getSigners()
 
     // 2
-    counter = await ethers.deployContract('Counter')
-
+    const gasPrice = 1637083528
+    const options: DeployContractOptions = { gasPrice }
+    counter = await ethers.deployContract('Counter', options)
     const initialCount = await counter.getCount()
 
     // 3
     expect(initialCount).to.eq(0)
     expect(await counter.getAddress()).to.properAddress
+    expect(counter.deploymentTransaction()?.gasPrice).to.eq(gasPrice)
   })
 
   // 4
--- a/packages/hardhat-test/typechain-types/hardhat.d.ts
+++ b/packages/hardhat-test/typechain-types/hardhat.d.ts
@@ -4,6 +4,7 @@
 
 import { ethers } from "ethers";
 import {
+  DeployContractOptions,
   FactoryOptions,
   HardhatEthersHelpers as HardhatEthersHelpersBase,
 } from "@nomicfoundation/hardhat-ethers/types";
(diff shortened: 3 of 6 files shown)


krzkaczor:
There is an example for ethers-v6 could you begin with reproducing the problem in that example. I guess just updating deps just suffice. Ie. we need some kind of regression check to make sure this doesn't break in the future.

DainisGorbunovs:
I've updated the following, and can confirm that the test passes:
1.  [examples/hardhat/test/counter.ts](https://github.com/dethcrypto/TypeChain/blob/5f8ac0376397512d69166b22f4e2a4877cea420e/examples/hardhat/test/counter.ts) test for a regression check
2. `@nomicfoundation/hardhat-ethers` package dependency version to 3.0.4



### Regression checks:
To ensure the written code in counter.ts works properly, I've checked how it behaves in the older hardhat-ethers version, and also if I avoid using type annotation. In these cases, either the test errored, or returned an invalid number (note `DeployContractOptions` allows us to override settings now).

<details>
<summary>if `@nomicfoundation/hardhat-ethers` remains at 3.0.0</summary>

```sh
An unexpected error occurred:

test/counter.ts:5:10 - error TS2724: '"@nomicfoundation/hardhat-ethers/types"' has no exported member named 'DeployContractOptions'. Did you mean 'deployContract'?

5 import { DeployContractOptions } from "@nomicfoundation/hardhat-ethers/types"
           ~~~~~~~~~~~~~~~~~~~~~

  ../../node_modules/.pnpm/@nomicfoundation+hardhat-ethers@3.0.0_ethers@6.3.0_hardhat@2.9.9/node_modules/@nomicfoundation/hardhat-ethers/types/index.d.ts:14:25
    14 export declare function deployContract(name: string, signerOrOptions?: ethers.Signer | FactoryOptions): Promise<ethers.Contract>;
                               ~~~~~~~~~~~~~~
    'deployContract' is declared here.

 ELIFECYCLE  Test failed. See above for more details.
```
</details>

<details>
<summary>same as above but also `DeployContractOptions` type annotation is not added</summary>

```sh
  Counter
    count up
      1) "before each" hook for "should count up"


  0 passing (300ms)
  1 failing

  1) Counter
       "before each" hook for "should count up":

      AssertionError: expected 1875000000 to equal 1637083528. The numerical values of the given "bigint" and "number" inputs were compared, and they differed.
      + expected - actual

      -1875000000
      +1637083528
      
      at /TypeChain/examples/hardhat/test/counter.ts:29:58
      at Generator.next (<anonymous>)
      at fulfilled (test/counter.ts:5:58)



 ELIFECYCLE  Test failed. See above for more details.
```

0xartcro (approved):

krzkaczor:
Superseded by: https://github.com/dethcrypto/TypeChain/pull/853

krzkaczor:
Thanks for this contribution! It will be released in a second.

PR #868, opened 2023-09-04
Normalize dirname

mymphe:
This PR fixes an issue where generating a variable name from a directory name starting with a digit causes the invalid syntax error (see pic below).

image

For instance, we organize Solidity files in directories by the version used, e.g. `0.4.24`, `0.8.9`.

## Naive solution
One simple solution would be to rename the directory to something like `v0.4.24`. However, we have verified our Solidity source code on Etherscan with this directory structure. If we rename the directories, we risk breaking our verification. This is why we can't use this approach.

## Our solution

We propose normalizing the generated variable name by prepending an underscore (`_`) to the name starting with a digit.
Thus, the dirname `0.4.24.` becomes `_0424` which is a legal JavaScript variable name. 

image


krzkaczor:
Thanks!

PR #869, opened 2023-09-15
barrel file moduleSuffix for type exports

leovigna:
When working with [tsx](https://github.com/esbuild-kit/tsx) I get issues with ESM interoperability due to the types not using explicity import paths. This PR fixes this and makes exports more consistent such that both types & implementations are exported with explicit `moduleSuffix` if specified.


leovigna:
Should I add a changeset. Please let me know if this PR is welcome.

PR #874, opened 2023-09-27
feat: add node16Modules flag for hardhat

raymondfeng:
This PR allows `@typechain/hardhat` to configure `node16Modules` flag so that we can generate ESM compatible code.


krzkaczor:
Thanks!

PR #887, opened 2023-12-08
fix: name clashes

### LINKED ISSUE #886: Generated types are broken if contract function names clash with ethers built-ins
**Example**: [This contract ](https://github.com/gnosis/zodiac/blob/master/contracts/core/Module.sol#L14) has a `target` property. The ethers v6 type generated by TypeChain looks like this:

```ts
/*
 * Interface 'Module' incorrectly extends interface 'BaseContract'.
 *  Types of property 'target' are incompatible.
 *   Type 'TypedContractMethod<[], [string], "view">' is not assignable to type 'string | Addressable'.ts(2430)                  
 */
export interface Module extends BaseContract {
  //             ~~~~~~ 
  // ...
  target: TypedContractMethod<[], [string], "view">;
  // ...
 }
```

**Expected behavior**: TypeChain should probably skip over functions that clash with ethers built-ins.

jfschwarz:
closes #886 

The ethers v6 `reservedKeywords` set was quite outdated. It had ethers v5 contract property names that no longer exist for v6 contracts (e.g.: `provider`) and was missing some v6 contract property names (e.g.: `target`). Rather than attempting to manually keep the `reservedKeywords` list up-to-date, I'd propose to directly read the list of built-in properties from the BaseContract class.

The `then` property is a peculiar case. It's not a property of BaseContract, but still won't be passed through by ethers as it is kept in a special [`passProperties`](https://github.com/ethers-io/ethers.js/blob/6ee1a5f8bb38ec31fa84c00aae7f091e1d3d6837/src.ts/contract/contract.ts#L640) list (see: [ethers v6 BaseContract Proxy](https://github.com/ethers-io/ethers.js/blob/6ee1a5f8bb38ec31fa84c00aae7f091e1d3d6837/src.ts/contract/contract.ts#L779)).


krzkaczor:
@jfschwarz thanks for this PR! Can you have a look at failing CI (`typecheck` task)?

jfschwarz:
Thank you, @krzkaczor! The failing CI typecheck tasks actually only reveal that ethers-v5 suffered under the same issue, so I went ahead and applied the fix also to that package.

Unfortunately, the dev tooling in that package appears a bit broken for me. I'm getting plenty of lint errors in target-ethers-v5  that don't seem to make make sense, not sure what's going on... I would appreciate if you could run the CI pipeline again to see if my latest changes check out.

jfschwarz:
Hey @krzkaczor, I finally got my local project tooling set up correctly and have now fixed the bug causing the tests to fail.

We have a few projects that need upgrading from Ethers 5 to 6, but we are currently blocked by the issue addressed in this PR. We understand that the project has been deprecated. However, we would greatly appreciate it if you could take a moment to review and release this fix, saving us from the need to fork TypeChain. 🙏 Let me know if I can support you in any way.

CJ42:
Is there any updates on this PR? Would be good to have this merged, as I also need this fix. It prevents compiling with nestjs

krzkaczor:
@CJ42 the project is not maintained anymore. Your best chance is if you want these changes, fork the project and publish to NPM yoursefl.

jfschwarz:
@CJ42 check if our fork works for you: https://github.com/gnosisguild/TypeChain

PR #898, opened 2024-04-01
Fix for barrel file codegen of type-only files

jmrossy:
Fix the import syntax for barrel file imports referring to type-only files. Fixes a problem with ESM compatibility when using the `--node16-modules` flag.

A follow-up to fix  #848

See related comments [here](https://github.com/dethcrypto/TypeChain/issues/819#issuecomment-2030453512) and [here](https://github.com/dethcrypto/TypeChain/issues/793#issuecomment-2030373004)

### DIFF
--- a/.changeset/hungry-noodle-dryer.md
+++ b/.changeset/hungry-noodle-dryer.md
@@ -1,0 +1,5 @@
+---
+'typechain': patch
+---
+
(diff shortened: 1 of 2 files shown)


nbayindirli (approved):
🚢

OKEAMAH (approved):
Run

